PR #1, opened 2017-07-07
DM-11039: Implement simple DIAObject

morriscb:


python/lsst/ap/association/dia_object.py
```diff
@@ -83,39 +97,34 @@ def __init__(self, dia_source_catalog, object_source_record=None):
...
             self.update()
         else:
             self._dia_object_record = object_source_record
-            self._generate_property_methods()
             self._updated = True
 
-    def _generate_property_methods(self):
-        """ Generate methods for accessing values stored in the
+    def get(self, name):
+        """ Return the data stored in column name within the internal
```
ctslater: If it's internal, then the user can't know about it.

tests/test_dia_object.py
```diff
@@ -20,28 +20,141 @@
...
+            Expected mean value of the parameters to be tested.
+        expected_std : float
+            Expected standard diviation of the DIAObject.
+
+        Return
+        ------
+        None
+        """
+
+        for name in self.dia_obj_schema.getNames():
```
ctslater: This really doesn't need to be a loop; you already know the names of the things you want to compare to.

python/lsst/ap/association/dia_object.py
```diff
@@ -55,12 +70,11 @@ class DIAObject(object):
         of DIASources changes and set to true when the initialize method is
         run.
     """
-
     def __init__(self, dia_source_catalog, object_source_record=None):
```
morriscb: This is for the case where you have already created a DIAObject and don't want to rerun the summary statistics on it's corresponding DIASourceCatalog. This would happen, for instance, when you persist DIAObjects and load them on a subsequent visit.

python/lsst/ap/association/dia_object.py
```diff
@@ -22,17 +22,32 @@
...
+    # so we add those.
+    schema.addField("coord_ra_rms", doc="rms position in ra/dec",
+                    type="Angle")
+    schema.addField("coord_dec_rms", doc="rms position in ra/dec",
+                    type="Angle")
+
+    return schema
 
 
 class DIAObject(object):
```
morriscb: Removed the described attributes.

python/lsst/ap/association/dia_object.py
morriscb: changed to stored.

morriscb: Current text changed to "Retrieve a specific summary statistic value from this DIAObject."

python/lsst/ap/association/dia_object.py
```diff
@@ -83,39 +97,34 @@ def __init__(self, dia_source_catalog, object_source_record=None):
...
+        Parameters
+        ----------
+        name : str or lsst.afw.table.Key
 
-        Returns
-        -------
-        None
+        Return
+        ------
+        An lsst.afw data type
```
morriscb: Changed to "A SourceRecord column value"

python/lsst/ap/association/dia_object.py
```diff
@@ -132,29 +141,51 @@ def update(self):
...
+        Returns
+        -------
+        None
         """
 
-        return self._updated
+        # Loop through DIASources, compute summary statistics (TBD) and store
+        # them in dia_object_record attribute.
+
+        for name in self.schema.getNames():
```
morriscb: Removed most of the functionality here and into it's own function _compute_mean_coordinate. Computation of errors has been moved to another ticket.

python/lsst/ap/association/dia_object.py
```diff
@@ -132,29 +141,51 @@ def update(self):
...
+            # For the MVP/S we are only dealing with angles so we skip over
+            # everything that isn't coord_ra or coord_dec.
+            if name != 'coord_ra' and name != 'coord_dec':
+                continue
+
+            mean_value = np.mean(self._dia_source_catalog[name])
```
morriscb: See previous comment. The computation also now uses afw.coord.averageCoord

python/lsst/ap/association/dia_object.py
```diff
@@ -132,29 +141,51 @@ def update(self):
...
+            # Currently hard coded to work with angles.
+            self._dia_object_record.set(name, afwGeom.Angle(mean_value))
+            # If we only have 1 source we can't compute an rms and hence we
+            # set the the rms variables to NaN.
+            if self.n_dia_sources > 1:
+                self._dia_object_record[name + '_rms'] = afwGeom.Angle(
+                    np.std(self._dia_source_catalog[name]))
```
morriscb: There is no method for compute a covariance matrix on a coordinate. See DM-7101

python/lsst/ap/association/dia_object.py
```diff
@@ -175,24 +206,10 @@ def append_dia_source(self, input_dia_source):
         # them until we are finished adding sources.
         self._updated = False
 
-        # Do stuff to append this SourceRecord.
+        self._dia_source_catalog.append(input_dia_source_record)
```
morriscb: Currently, this does not copy the source. This will be taken care of by the update method later which deep copies the full catalog. Would it be enough to add a comment to the docstring that the source isn't properly copied until update is run?

python/lsst/ap/association/dia_object.py
```diff
@@ -231,6 +262,16 @@ def dia_source_catalog(self):
         return self._dia_source_catalog
 
     @property
+    def n_dia_sources(self):
```
morriscb: Perhaps but I thought it would be useful to be explicit as to which length was being returned. __len__ could be interpreted as returning a property of the the DIAObject not necessarily the number of dia_sources in this object.

tests/test_dia_object.py
```diff
@@ -20,28 +20,141 @@
...
-    def compute_light_curve(self):
-        pass
 
-    def compute_summary_statistics(self):
+class TestDIAObject(unittest.TestCase):
+
+    def setUp(self):
+        """ Create the DIAObjct schema we will use throughout these tests.
+        """
+        self.dia_obj_schema = make_minimal_dia_object_schema()
```
morriscb: Correct. This was left over from when before I had the create_test_dia_sources method in the test. I have removed the setUp and tearDown methds.

tests/test_dia_object.py
```diff
@@ -20,28 +20,141 @@
...
+
+    def test_update(self):
+        """ If we instantiate the class with a set of DIASources we test to
+        make sure that the DIAObject is intantiated correctly and that we
+        compute the summary statistics in the expected way.
+        """
+        sources = create_test_dia_sources(5)
+        dia_obj = DIAObject(sources, None)
+
+        self._compare_dia_object_values(dia_obj, 0, 2.0, 1.4142135623730951)
```
morriscb: The idea is that the DIAObjects's internal schema will change as more summary statistics get added beyond the MVP/S. This function made writing the comparisons in one place for all the tests easier.

morriscb:
Committed updated version with incorporated reviewer responses.

python/lsst/ap/association/dia_object.py
```diff
@@ -22,45 +22,47 @@
...
-        of DIASources changes and set to true when the initialize method is
-        run.
     """
-
     def __init__(self, dia_source_catalog, object_source_record=None):
         """  Create a DIAObject given an input SourceCatalog of
         DIASources.
 
-        Takes as input an lsst.afw.table.SourceCatalog object specifying a
+        Takes as input an lsst.afw.table.SourceCaatlog object specifying a
```
ctslater: Typo.

python/lsst/ap/association/dia_object.py
```diff
@@ -22,45 +22,47 @@
...
     """ Define and create the minimal schema required for a DIAObject.
 
     Return
     ------
     lsst.afw.table.schema.schema.Schema
     """
-    Raise(NotImplementedError)
+
+    schema = afwTable.SourceTable.makeMinimalSchema()
+    # Currently we do not store
```
ctslater: Broken sentence

python/lsst/ap/association/dia_object.py
```diff
@@ -22,45 +22,47 @@
...
+    schema = afwTable.SourceTable.makeMinimalSchema()
+    # Currently we do not store 
+
+    # For the MVP/S we currently only care about the position though
+    # in the future we will add summary computations for fluxes etc.
+    # as well as their errors.
+
+    # Currently the covariance of a coordinate is not defined in the
+    # stack, so we will hold off until it is implemented. This is to be
+    # addressed in DM-7101.
```
ctslater: These comments only make sense if one knows that fields for the ra/dec uncertainty were removed. I'd drop them.

python/lsst/ap/association/dia_object.py
```diff
@@ -83,78 +85,80 @@ def __init__(self, dia_source_catalog, object_source_record=None):
...
+    def _compute_mean_coordinate(self):
+        """ Compute the mean coordinate of this DIAObject given the current
+        DIASources associated with it.
         """
 
-        return self._updated
+        coord_list = [src.getCoord() for src in self._dia_source_catalog]
+        ave_coord = averageCoord(coord_list)
+        return ave_coord
+        self._dia_object_record.setCoord(ave_coord)
```
ctslater: Broken; unreachable code.

python/lsst/ap/association/dia_object.py
```diff
@@ -203,10 +187,24 @@ def get_light_curve(self):
         """
 
         # Loop through DIASources and return the "light curve"
-        # I'm keeping this separate from coupte summary statistics for the
-        # moment.
+        # Right now I'm making this the same as returning the
+        # dia_source_catalog.
+
+        return self.dia_source_catalog()
```
ctslater: This should raise `NotImplementedError` rather than returning something that you later plan on changing.

tests/test_dia_object.py
ctslater: I don't think that actually makes it easier; say you now want to compare a mean flux, you'll still have to go to every instance of `_compare_dia_object_values` and change them. At least make these keyword arguments so they're not just anonymous numbers laying around.

tests/test_dia_object.py
```diff
@@ -20,28 +20,125 @@
...
+        """
+        single_source = create_test_dia_sources(1)
+
+        dia_obj = DIAObject(single_source, None)
+        dia_obj_dup = DIAObject(single_source, dia_obj.dia_object_record)
+
+        self._compare_dia_object_values(dia_obj, 0, 0.0, np.nan)
+        self._compare_dia_object_values(dia_obj_dup, dia_obj.id, 0.0, np.nan)
+
+    def _compare_dia_object_values(self, dia_object, expected_id,
```
ctslater: This function is so much nicer now. `expected_std` should be eliminated since it's unused.

tests/test_dia_object.py
ctslater: These functions are still here ...?

morriscb: Must not have been committed. They are not removed.

python/lsst/ap/association/dia_object.py
morriscb: Clarified this to say "In the future we would like to use store the convariance but,... etc."

PR #2, opened 2017-07-27
tickets/DM-11283: Spec out DIAObjectCollection API

morriscb:


python/lsst/ap/association/dia_collection.py
```diff
@@ -19,3 +19,188 @@
...
+    def __init__(self, dia_objects):
+        """ Initialize a collection of dia_objects.
+
+        Store and update a list of dia_objects in the collection.
+
+        Parameters
+        ----------
+        dia_objects : a list of DIAObjects
+            List of DIAObjects that represent the collection of variable
+            objects for this visit. Each DIAOjbect will be individually
```
ebellm: DIAOjbect -> DIAObject

python/lsst/ap/association/dia_collection.py
```diff
@@ -19,3 +19,188 @@
...
+
+        Returns
+        -------
+        (N DIAObject by M DIASource) float array of between DIAObjects and
+        DIASources.
+        """
+        pass
+
+    def match(self, dia_source_catalog, scores):
+        """ Append DIAsources to DIAObjectds given a score and create new
```
ebellm: DIAObjectds -> DIAObjects

python/lsst/ap/association/dia_collection.py
```diff
@@ -19,3 +19,188 @@
...
+        DIAObjects in this collection from DIASources with poor scores.
+
+        Parameters
+        ----------
+        dia_source_catalog : an lsst.afw.SourceCatalog
+            A contiguous catalo of dia_sources for which the set of scores
```
ebellm: catalo -> catalog

python/lsst/ap/association/dia_collection.py
```diff
@@ -19,3 +19,188 @@
...
+        between this collection and an input diat_source catalog.
+
+        max_dist sets maximum seperation in arcseconds to consider a
+        dia_source a possible match to a dia_object. If the pair is
+        beyond this distance no score is computed.
+
+        Parameters
+        ----------
+        dia_source_catalog : an lsst.afw.SourceCatalog
+            A contiguous catalo of dia_sources to "score" based on distance
```
ebellm: catalo -> catalog

python/lsst/ap/association/dia_collection.py
```diff
@@ -19,3 +19,188 @@
...
+
+
+class DIAObjectCollection(object):
+    """ A collection of DIAObjects with convenience functions for scoring and
+    matching DIASources into the collection of DIAObjects.
+
+    Attributes
+    ----------
+    dia_objects : a list of DIAObjects
+        List of DIAObjects representing this collection the current (e.g.)
```
ebellm: Rather than storing the DIAObjects as a list, I think it may make sense to store them in a dictionary for fast access by id.  I am open to counterarguments, though.

python/lsst/ap/association/dia_collection.py
```diff
@@ -19,3 +19,188 @@
...
+
+        If the tree has not been updated with the current positions of
+        all DIAObjects internal to this collection we return false.
+
+        Return
+        ------
+        bool
+        """
+
+        return self._is_valid_tree
```
ebellm: I don't see any methods dealing with persistence to/from a data store in here.  Should those be internal to the class or external?  My guess is external (mocking up something like Butler.put()), etc., but can you tell me your thoughts here?

python/lsst/ap/association/dia_collection.py
morriscb: That makes sense. One hesitation I have for this is that the scoring and matching step is most easily done on the relative position in a list rather than a id in a dictionary. What if I add a dictionary internal to the class that is keyed on the id and returns the index in the list? I think I would hide this from user through some overwriting of built in functions or create a accessor method that uses the id as the input and returns the DIAObject of interest through the dictionary of id to list position.

python/lsst/ap/association/dia_collection.py
morriscb: Sure. I think this is something that gets taken care of by the future AssociationTask. On instantiation this would query where ever the DIAObjects are stored and then create the DIAObjectCollection for the association step. AssociationTask would then take care of writing the data as well.

python/lsst/ap/association/dia_collection.py
morriscb: Just made a first past commit on a the dictionary I described. Let me know if this works for you.

ebellm: Looks great!  Ship it.

PR #3, opened 2017-08-07
DM-11041: Implement simple DIAObject to DIASource matching algorithm and flesh out DIACollection

morriscb:


tests/test_dia_collection.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def test_append_and_update(self):
+        """ Test that we can add a new DIAObject to an existing
+        DIAObjectCollection.
+        """
+        obj_list = create_test_dia_objects(1, 1, 0.1)
+        obj_collection = DIAObjectCollection(obj_list)
+
+        new_dia_obj = create_test_dia_objects(1, 1, 0.1, 1, 0.1)[0]
+        obj_collection.append(new_dia_obj)
```
parejkoj: If you use `extend` here, you don't have to take the first element above.

tests/test_dia_collection.py
```diff
@@ -0,0 +1,172 @@
...
+        self.assertFalse(obj_collection.is_updated)
+        self.assertFalse(obj_collection.is_valid_tree)
+
+        obj_collection.update_dia_objects()
+        self.assertTrue(obj_collection.is_updated)
```
parejkoj: Put another `self.assertFalse(obj_collection.is_valid_tree)` right here?

tests/test_dia_collection.py
```diff
@@ -0,0 +1,172 @@
...
+        obj_collection.update_spatial_tree()
+        self.assertTrue(obj_collection.is_valid_tree)
+
+        self.assertEqual(len(obj_collection.get_dia_object_ids()), 2)
```
parejkoj: These two lines should just be `self.assertEqual(obj_collection.get_dia_object_ids(), [1,2])`, no?

tests/test_dia_collection.py
```diff
@@ -0,0 +1,172 @@
...
+        This also tests that a DIASource that can't be associated within
+        tolerance is addpended to the DIAObjectCollection as a new
+        DIAObject.
+        """
+        obj_collection = DIAObjectCollection(
+            create_test_dia_objects(4, 1, -1))
+        src_cat = create_test_dia_sources(5)
+        for src_idx, src in enumerate(src_cat):
+            edit_source_record(
+                src, src_idx + 4, 0.1 * src_idx, 0.1 * src_idx, -1)
```
parejkoj: Again, if you're going to split the args on new lines, do them all (and you shouldn't have to for this one)

tests/test_dia_collection.py
```diff
@@ -0,0 +1,172 @@
...
+
+        self.assertEqual(len(obj_collection.get_dia_object_ids()), 2)
+        self.assertEqual(obj_collection.get_dia_object(1).get('id'), 1)
+
+    def test_score_and_match(self):
+        """ Test association between a set of sources and an existing
+        DIAObjectCollection.
+
+        This also tests that a DIASource that can't be associated within
+        tolerance is addpended to the DIAObjectCollection as a new
```
parejkoj: addpended?

tests/test_dia_collection.py
```diff
@@ -0,0 +1,172 @@
...
+        src_cat = create_test_dia_sources(5)
+        for src_idx, src in enumerate(src_cat):
+            edit_source_record(
+                src, src_idx + 4, 0.1 * src_idx, 0.1 * src_idx, -1)
+        score_struct = obj_collection.score(
+            src_cat,  afwGeom.Angle(1.0, units=afwGeom.arcseconds))
+
+        self.assertFalse(np.isfinite(score_struct.scores[-1]))
+        for src_idx in range(4):
+            self.assertAlmostEqual(score_struct.scores[src_idx], 0.0,
```
parejkoj: Why shouldn't they be identically zero?

tests/test_dia_collection.py
```diff
@@ -0,0 +1,172 @@
...
+                                   places=9)
+
+        updated_indices = obj_collection.match(src_cat, score_struct)
+        self.assertEqual(len(obj_collection.dia_objects), 5)
+
+        for idx, obj_id in enumerate(obj_collection.get_dia_object_ids()):
+            self.assertEqual(idx, updated_indices[idx])
+            if idx != 4:
```
parejkoj: Why is `idx==4` different?

tests/test_dia_collection.py
```diff
@@ -0,0 +1,172 @@
...
+
+class MemoryTester(lsst.utils.tests.MemoryTestCase):
+    pass
+
+
+def setup_module(module):
+    lsst.utils.tests.init()
+
+if __name__ == "__main__":
+
```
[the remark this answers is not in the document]
timj: Also not PEP8 compliant.

timj:
There are quite a few PEP8 violations in the file. Maybe some of them were always there but given the amount of code changing can we fix them? And since we are fixing them, please add a `setup.cfg` to enable flake8 tests and enable test auto discovery.

timj:
There are quite a few PEP8 violations in the file. Maybe some of them were always there but given the amount of code changing can we fix them? And since we are fixing them, please add a `setup.cfg` to enable flake8 tests and enable test auto discovery.

python/lsst/ap/association/dia_collection.py
```diff
@@ -22,9 +22,13 @@
...
 
-from scipy.spatial import cKDtree
+import numpy as np
+from scipy.spatial import cKDTree
 
-import .dia_object
+import lsst.afw.table as afwTable
+import lsst.pipe.base as pipeBase
+
+from .dia_object import *
```
timj: Usually we don't allow wildcard imports into a non-`__init__`.py.

timj:
There are at least two real bugs found by flake8:
```
./python/lsst/ap/association/dia_collection.py:268:52: E712 comparison to False should be 'if cond is False:' or 'if not cond:'
./python/lsst/ap/association/dia_object.py:192:15: F821 undefined name 'NotImplimentedError'
```

python/lsst/ap/association/dia_collection.py
```diff
@@ -168,18 +195,35 @@ def score(self, dia_source_catalog, max_dist_arcsec):
...
+        lsst.pipe.base.struct
+            Contains two arrays: scores DIAObjects, indices of DIAObject
+            mached to.
         """
-        pass
+        if not self._is_valid_tree:
+            self.update_spatial_tree()
+
+        scores = np.ones(len(dia_source_catalog)) * np.nan
+        obj_indices = (np.ones(len(dia_source_catalog), dtype=np.int) *
```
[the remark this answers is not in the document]
timj: I think so that it can break over a new line without a backslash.

python/lsst/ap/association/association.py
```diff
@@ -0,0 +1,34 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+
```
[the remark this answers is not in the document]
morriscb: For my future reference this is something that should be included in every file yes?

python/lsst/ap/association/association.py
```diff
@@ -0,0 +1,34 @@
...
+__all__ = ["AssociationConfig", "AssociationTask"]
+
+import lsst.pex.config as pexConfig
+import lsst.pipe.base as pipeBase
+
+class AssociationConfig(pexConfig.Config):
+    pass
+
+
+class AssociationTask(pipeBase.Task):
```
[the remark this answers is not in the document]
morriscb: For now it is just a stub. The ticket to complete the class is DM-11747.

python/lsst/ap/association/dia_collection.py
```diff
@@ -115,24 +133,33 @@ def update_dia_objects(self, force=False):
                 dia_object.update()
                 self._is_valid_tree = False
 
-        self._is_updated
+        self._is_updated = True
 
-        return None
+        return self._is_updated
```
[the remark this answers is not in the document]
morriscb: Yes it is. I want this variable to reflect the whether this method was run or not.

python/lsst/ap/association/dia_collection.py
```diff
@@ -188,15 +232,51 @@ def match(self, dia_source_catalog, scores):
...
-        pass
+
+        used_dia_object = np.zeros(len(self.dia_objects), dtype=np.bool)
+        used_dia_source = np.zeros(len(dia_source_catalog), dtype=np.bool)
+
+        n_previous_objects = len(self.dia_objects)
+
+        updated_and_new_dia_objects = []
+
+        score_args = score_struct.scores.argsort(axis=None)
```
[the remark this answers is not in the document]
morriscb: Comment added.

python/lsst/ap/association/dia_collection.py
```diff
@@ -188,15 +232,51 @@ def match(self, dia_source_catalog, scores):
...
+        for score_idx in score_args:
+            if not np.isfinite(score_struct.scores[score_idx]):
+                break
```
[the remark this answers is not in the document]
morriscb: It should be break. Because of the sort all of the values past the first NaN are also NaN so we don't need to loop over them. I have added a comment to reflect this.

python/lsst/ap/association/dia_collection.py
```diff
@@ -188,15 +232,51 @@ def match(self, dia_source_catalog, scores):
...
+            used_dia_source[score_idx] = True
+            updated_and_new_dia_objects.append(
+                score_struct.indices[score_idx])
+
+            self.dia_objects[
+                score_struct.indices[score_idx]].append_dia_source(
+                    dia_source_catalog[score_idx])
+
+        n_new_objects = 0
+        for src_idx in np.argwhere(used_dia_source == False).flatten():
```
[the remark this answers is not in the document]
morriscb: numpy argwhere returns, in this case, an array of shape (N, 1) instead of (N,). This is the reason for the flatten. I've removed the flatten and written

    for (src_idx,) in np.argwhere(used_dia_source == False):

hopefully this is more clear.

tests/test_dia_collection.py
morriscb: I've changed this to len(dia_objects) - 1 and added a comment explaining why this index is special.

tests/test_dia_collection.py
morriscb: They should be. I'm being a bit cautious here as I don't want to assert that two floats be identically equal. If I don't need to be I am happy to change it.

tests/test_dia_collection.py
morriscb: dia_collection here is not a list object and does not have an extend method. Are you suggesting implementing such a method within DIAObjectCollection?

tests/test_dia_collection.py
```diff
@@ -0,0 +1,172 @@
...
+    """ Create DIAObjects with a specified number of DIASources attached.
+
+    Parameters
+    ----------
+    n_objects : int
+        Number of DIAObjects to generate.
+    n_src : int
+        Number of DIASources to generate for each DIAObject.
+    scatter_arcsec : float
+        Scatter to add to the position of each DIASource.
```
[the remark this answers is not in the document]
morriscb: Doc strings added. I'm a bit hesitant to make these Angles as I do just use the float random number generator from numpy for the scatter. It makes a bit more sense to me to just add the floats up and convert to a Angle in the editing of the coordinate for the sources.

tests/test_dia_collection.py
```diff
@@ -0,0 +1,172 @@
...
+    Returns
+    -------
+    A list of DIAObjects
+    """
+    output_dia_objects = []
+    for obj_idx in range(n_objects):
+        src_cat = create_test_dia_sources(n_src)
+        for src_idx in range(n_src):
+            edit_source_record(
+                src_cat[src_idx], start_id + n_src * obj_idx + src_idx,
```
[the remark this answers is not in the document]
morriscb: Moved the statements to one line per variable.

tests/test_dia_collection.py
```diff
@@ -0,0 +1,172 @@
...
+                start_angle_degrees + 0.1 * obj_idx,
+                start_angle_degrees + 0.1 * obj_idx,
+                scatter_arcsec)
+        output_dia_objects.append(DIAObject(src_cat))
+    return output_dia_objects
+
+
+def edit_source_record(src, src_id, ra_degrees, dec_degrees, scatter_arcsec):
+    """ Edit the center coordinate and id of a source record in place.
```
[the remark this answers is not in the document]
morriscb: Would you accept "edit_and_scatter_source_record"?

tests/test_dia_collection.py
```diff
@@ -0,0 +1,172 @@
...
+    id : int
+        Integer value to set the record id to.
+    ra_degrees : float
+        RA value to set the record coordinate to.
+    dec_degrees : float
+        DEC value to set the record coordinate to.
+    scatter_arcsec : float
+        Arcsecond scatter to add to the position of the source record coord.
+    """
+    coord = Coord(afwGeom.Angle(ra_degrees, units=afwGeom.degrees),
```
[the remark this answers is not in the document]
morriscb: I'm a bit okay with this as is. As I stated previously it feels a bit awkward to input an Angle only to get the float to use with numpy.random and then convert back to an Angle.

python/lsst/ap/association/dia_collection.py
morriscb: Also, I don't quite know what you mean be "setup.cfg", Tim. Is this a file I should add to the repository somewhere? If so is there an example you can link me to, please?

python/lsst/ap/association/association.py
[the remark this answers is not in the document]
parejkoj: Yes, module-level docstrings should always exist, even if they are just a sentence:

https://developer.lsst.io/docs/py_docs.html#modules

python/lsst/ap/association/association.py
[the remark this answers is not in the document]
parejkoj: Put that that ticket as a TODO in the class docstring then.

python/lsst/ap/association/dia_collection.py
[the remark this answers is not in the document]
parejkoj: Ah, I see. A comment to that effect would be good.

python/lsst/ap/association/dia_collection.py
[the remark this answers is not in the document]
parejkoj: Oh, interesting. I think that's worth a very short comment.

tests/test_dia_collection.py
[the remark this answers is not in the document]
parejkoj: That's fair on the angles. At least the names make it clear, and it's only for the tests.

tests/test_dia_collection.py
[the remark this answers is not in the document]
parejkoj: Sure, that works.

tests/test_dia_collection.py
parejkoj: Oh, I thought it was a list. Oh well!

python/lsst/ap/association/dia_collection.py
```diff
@@ -168,18 +197,38 @@ def score(self, dia_source_catalog, max_dist_arcsec):
...
+            * scores: array of floats of match quality
+            * indices: index in DIAObjectCollection that source matched to
+            Default values for these arrays are NaN and the number of
+            DIAObjects in this collection, respectively.
         """
-        pass
+        if not self._is_valid_tree:
+            self.update_spatial_tree()
+
+        scores = np.ones(len(dia_source_catalog)) * np.nan
```
parejkoj: If you're just making an array of NaNs, use `np.empty`, since you don't need to initialize it.

morriscb: I need the the NaNs to have a positive sign so that they end up at the end of a sort. I'd like to keep this then, unless there is a way to do this with np.empty

parejkoj: Per our conversation, NaN isn't signed, and what you really want is `np.inf`(or the negative thereof).

tests/test_dia_collection.py
parejkoj: It's up to you: if you expect that the above items really should have a score of zero, then use exact equality. If you think they should have "very small" scores, then leave it as is.

python/lsst/ap/association/dia_collection.py
parejkoj: @morriscb your reply is to the wrong comment of @timj : I think you meant to reply to his comment just above, re: adding setup.cfg.

Unfortunately, @timj it looks like DM-10493 was closed Won't Fix, so I don't know where the documentation for configuring auto-flake8 actually lives...

timj: It would be great if you could enable flake8 testing in this package. You do that by copying a `setup.cfg` from another package (until I update the dev guide). Something like this is fine:
```
[flake8]
max-line-length = 110
ignore = E133, E226, E228, N802, N803, N806
exclude = __init__.py

[tool:pytest]
addopts = --flake8
flake8-ignore = E133 E226 E228 N802 N803 N806
```
Then you enable pytest auto discovery in `tests/SConscript` by adding `pyList=[]` to the `tests()` method.

tests/test_dia_collection.py
morriscb: I'll change it to identical then.

python/lsst/ap/association/dia_collection.py
morriscb: Hi @timj, I tried this and I get "scons: Nothing to be done for `tests'." when using pyList=[] in the tests() method. When I remove this, scons then performs the tests. Could you take a look at the current version that I committed, please, and let me know what is missing? I Copied your text above and put it into setup.cfg in the root directory of the repository.

tests/test_dia_collection.py
morriscb: After testing this, the value is not identically zero (it's of order 1e-19). I've changed the test back to assertAlmostEqual(value1, value2, places=16). I've also added a comment stating that this value should be really close to 0.0 but not identical.

python/lsst/ap/association/dia_collection.py
timj: Are you running with an old weekly? You need current master or weekly _35.

morriscb: Thanks @timj, everything works now and comes up "green".

PR #7, opened 2017-11-14
DM-11807: Create initial DIAObject summary statistics for ap_verify

morriscb:


python/lsst/ap/association/dia_object.py
```diff
@@ -190,9 +198,27 @@ def append_dia_source(self, input_dia_source_record):
         # them until we are finished adding sources.
         self._updated = False
 
-        self._dia_source_catalog.append(
-            self._dia_source_catalog.getTable().copyRecord(
-                input_dia_source_record))
+        # We need to test that schema of the source to appended lines up with
```
parejkoj: "to be appended"

python/lsst/ap/association/dia_object.py
```diff
@@ -190,9 +198,27 @@ def append_dia_source(self, input_dia_source_record):
...
+        # assuming that the schema defined in this dia_object is a subset of
+        # of the input source's schema.
+        input_schema = input_dia_source_record.getSchema()
+        if input_schema == self._dia_source_schema:
+            self._dia_source_catalog.append(
+                self._dia_source_catalog.getTable().copyRecord(
+                    input_dia_source_record))
+        else:
+            tmp_source_record = afwTable.SourceTable.makeRecord(
+                afwTable.SourceTable.make(self._dia_source_schema))
```
parejkoj: Why is this `makeRecord(make())`? Can't you just pass the schema to `makeRecord`?

python/lsst/ap/association/dia_object.py
```diff
@@ -190,9 +198,27 @@ def append_dia_source(self, input_dia_source_record):
...
-        self._dia_source_catalog.append(
-            self._dia_source_catalog.getTable().copyRecord(
-                input_dia_source_record))
+        # We need to test that schema of the source to appended lines up with
+        # the schema specified in the catalog of dia_sources associated with
+        # this dia_object. If not we attempt to access the those that overlap
+        # assuming that the schema defined in this dia_object is a subset of
+        # of the input source's schema.
+        input_schema = input_dia_source_record.getSchema()
+        if input_schema == self._dia_source_schema:
```
parejkoj: This `if:else:` block seems overly complicated. Is there not a simpler way to add the record?

tests/test_dia_object.py
```diff
@@ -42,6 +42,9 @@ def create_test_dia_sources(n_sources=5):
     -------
     A lsst.afw.SourceCatalog
     """
+    if schema is None:
+        schema = schema
```
parejkoj: What's going on here? This line makes no sense.

tests/test_dia_object.py
```diff
@@ -144,6 +147,40 @@ def test_dia_source_append_and_update(self):
         dia_obj.update()
         self._compare_dia_object_values(dia_obj, 0, compare_values)
 
+    def test_dia_source_append_different_schema(self):
```
parejkoj: Does the above test test the "same schema" case, or do you need a different test for that?

tests/test_dia_object.py
```diff
@@ -144,6 +147,40 @@ def test_dia_source_append_and_update(self):
...
+        associated_sources = dia_obj.dia_source_catalog
+        self.assertEqual(len(associated_sources), 2)
+        self.assertEqual(associated_sources[-1].getId(),
+                         sources[-1].getId())
+        self.assertEqual(associated_sources[-1].getCoord(),
+                         sources[-1].getCoord())
+
+        compare_values = {
+            "coord_ra": 0.4999619199226218,
+            "coord_dec": 0.5000190382261059
```
parejkoj: Where do these numbers come from?

tests/test_dia_object.py
```diff
@@ -144,6 +147,40 @@ def test_dia_source_append_and_update(self):
...
+    def test_dia_source_append_different_schema(self):
+        """ Test the appending of a DIASource to a DIAObject where the DIASource
+        has a more complex schema.
+
+        We also test that the update function works properly and that the
+        summary statistics are computed as expected.
+        """
+        single_source = create_test_dia_sources(1)
+        dia_obj = DIAObject(single_source, None)
+        self.assertEqual(dia_obj.id, single_source[0].getId())
```
parejkoj: Can you do `self.assertEqual(dia_obj, single_source[0])` to test all the fields?

tests/test_dia_object.py
morriscb: Oops, I wanted this to be make_minimal_dia_source_schema(). It is now fixed.

tests/test_dia_object.py
morriscb: It does test the same schema case.

tests/test_dia_object.py
morriscb: No. I'm testing here that some of the more important values have been copied into the dia_object which is not the same data structure as the source record I access from the source catalog.

tests/test_dia_object.py
morriscb: I've added a comment. Basically it's just average position of the incremented DIASources that are put into this object.

parejkoj: It would be better if you actually computed that average: these are too much "magic value" for my tastes, given that they depend on code that lives elsewhere in this file.

python/lsst/ap/association/dia_object.py
morriscb: You are correct. Since I have a SourceCatalog available I can use the makeRecord method on that to create the empty record object. If I didn't have the catalog sitting around and wanted to create a SourceRecord I would need to first create the SourceTable from the schema which is what this line does.

python/lsst/ap/association/dia_object.py
morriscb: Its complicated because I want to get a copy of the record as well as make sure the schemas line up. The way to do the second could possibly be taken care of by a SchemaMapper but creating the mapping may in fact result in more complex code than what you see here.

tests/test_dia_object.py
morriscb: Okay, I'll use the original source catalog arrays to compute the values to get rid the magic numbers.

morriscb:
Committed first round of responses to review.

PR #8, opened 2017-11-15
DM-12538: Cut loading of DIAObjects to the ccd BoundingBox area in ap_assocation

morriscb:


python/lsst/ap/association/assoc_db_sqlite.py
```diff
@@ -377,6 +385,10 @@ def _get_dia_objects(self, indexer_indices):
         for row in self._query_dia_objects(indexer_indices):
             dia_object_record = \
                 self._dia_object_converter.source_record_from_db_row(row)
+            if expMd is not None and \
+               not self._check_dia_object_possition(dia_object_record,
```
isullivan: "position" not "possition"

python/lsst/ap/association/assoc_db_sqlite.py
```diff
@@ -419,6 +431,27 @@ def _query_dia_objects(self, indexer_indices):
 
         return output_rows
 
+    def _check_dia_object_possition(self, dia_object_record, bbox, wcs):
```
isullivan: One "s" in position

python/lsst/ap/association/assoc_db_sqlite.py
```diff
@@ -377,6 +385,10 @@ def _get_dia_objects(self, indexer_indices):
         for row in self._query_dia_objects(indexer_indices):
             dia_object_record = \
                 self._dia_object_converter.source_record_from_db_row(row)
+            if expMd is not None and \
+               not self._check_dia_object_possition(dia_object_record,
+                                                    expMd.bbox, expMd.wcs):
+                continue
```
isullivan: This is difficult to read. A comment would help, but you might also consider renaming `_check_dia_object_position` and have it return `True` if the source is to be excluded. That would allow you to remove the `not` in the if statement, which would make it easier to read.

python/lsst/ap/association/assoc_db_sqlite.py
morriscb: I've moved the "is not None" check into the function and changed the next lines to be within the if statement so that now only "if self._check_dia_object_position" is checked.

python/lsst/ap/association/assoc_db_sqlite.py
morriscb: Oops I put this comment in the wrong place so I'll just put it here too: I've moved the "is not None" check into the function and changed the next lines to be within the if statement so that now only "if self._check_dia_object_position" is checked.

PR #9, opened 2017-11-30
DM-11155: Implement object count metrics

morriscb:


python/lsst/ap/association/association.py
```diff
@@ -118,9 +118,25 @@ def associate_sources(self, dia_collection, dia_sources):
...
+
+        self._add_association_meta_data(match_result)
 
         return pipeBase.Struct(
             dia_collection=dia_collection,
-            updated_ids=updated_dia_objects
+            updated_ids=match_result.updated_and_new_dia_object_ids,
         )
+
+    def _add_association_meta_data(self, match_result):
```
isullivan: Is this metadata used anywhere, or is it for future development?

morriscb: It's used in ap_verify. This is a stop gap until the data products AssocationTask produces better routed through the Butler or similar system.

PR #10, opened 2017-12-18
DM-13052: Association does not call update on associated DIAObjects

morriscb:


morriscb:
Technically the DPDD just lists that DIAObject RA DEC will be the observation at some time with the time of this observation stored. To me right now the last detection RA DEC makes sense to capture the shift of slowly moving objects. I don't think averaging versus not will make much difference in right now as the code does not currently compute or use an error ellipse in RA DEC to catch objects that may be faint and have large deviations in centroiding.

SimonKrughoff:
OK.  Fine by me.

PR #11, opened 2018-02-21
DM-13477: Move association math from DIAObjectCollection into AssociationTask

morriscb:
Moved the score and match methods from DIAObjectCollection into AssociationTask. Also fixed problems associated with SkyWcs.

### DIFF
--- a/python/lsst/ap/association/association.py
+++ b/python/lsst/ap/association/association.py
@@ -25,10 +25,14 @@
 
 from __future__ import absolute_import, division, print_function
 
+import numpy as np
+
 import lsst.pex.config as pexConfig
 import lsst.pipe.base as pipeBase
 import lsst.afw.geom as afwGeom
+import lsst.afw.table as afwTable
 from .assoc_db_sqlite import AssociationDBSqliteTask
+from .dia_object import DIAObject
 
 __all__ = ["AssociationConfig", "AssociationTask"]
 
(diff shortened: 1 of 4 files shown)


python/lsst/ap/association/association.py
```diff
@@ -117,10 +121,16 @@ def associate_sources(self, dia_collection, dia_sources):
             DIASources into.
         dia_sources : lsst.afw.table.SourceCatalog
             DIASources to associate into the DIAObjectCollection.
+
+        Returns
+        -------
+        lsst.pipe.base.Struc
```
r-owen: Please document what fields the struct contains

Also `Struc` -> `Struct`

tests/test_association_task.py
```diff
@@ -344,6 +344,68 @@ def test_associate_sources(self):
...
+        for updated_idx, obj_id in enumerate(updated_ids):
+            if updated_idx == len(updated_ids) - 1:
+                # We created a new DIAObject in the collection hence the last
+                # DIAObject in this collection is new and contains only one
+                # DIASource.
+                self.assertEqual(
+                    dia_collection.get_dia_object(obj_id).n_dia_sources, 1)
+            else:
+                self.assertEqual(
+                    dia_collection.get_dia_object(obj_id).n_dia_sources, 2)
```
r-owen: What you have works, but I feel the following is a bit easier to read. Your choice.
```
        # We created a new DIAObject in the collection hence the last
        # DIAObject in this collection is new and contains only one
        # DIASource. All others contain two.
        self.assertEqual(
            dia_collection.get_dia_object(updated_ids[-1]).n_dia_sources, 1)
        for obj_id in updated_ids[0:-1]:
            self.assertEqual(
                dia_collection.get_dia_object(obj_id).n_dia_sources, 2)
```

PR #12, opened 2018-03-16
DM-11162: Replace all use of Coord and subclasses with SpherePoint

r-owen:


.gitignore
```diff
@@ -1,6 +1,7 @@
+.cache
 .sconsign.dblite
-config.log
 .sconf_temp
+config.log
```
rearmstr: Is there a reason to change where ```config.log``` appears in the file?

python/lsst/ap/association/assoc_db_sqlite.py
```diff
@@ -136,7 +136,7 @@ def source_record_to_value_list(self, source_record):
 
         Parameters
         ----------
-        source_record : afw.table.SourceRecord
+        source_record : lsst.afw.table.SourceRecord
```
rearmstr: Not related to the Coord -> SpherePoint conversion.

.gitignore
r-owen: I prefer having the .scons* files together

python/lsst/ap/association/assoc_db_sqlite.py
r-owen: True. Are you requesting that it be on a separate commit?

tests/test_dia_collection.py
```diff
@@ -90,13 +89,11 @@ def edit_and_offset_source_record(src, src_id, ra_degrees, dec_degrees,
     scatter_arcsec : float
         Arcsecond scatter to add to the position of the source record coord.
     """
-    coord = Coord(afwGeom.Angle(ra_degrees, units=afwGeom.degrees),
-                  afwGeom.Angle(dec_degrees, units=afwGeom.degrees))
+    coord = lsst.afw.geom.SpherePoint(ra_degrees, dec_degrees, afwGeom.degrees)
```
mrawls: This is a lovely example of how SpherePoint makes our lives better! 💯

python/lsst/ap/association/assoc_db_sqlite.py
rearmstr: I'm fine with it, but do we have a policy on this?  I've been asked by other reviewers to do cleanups in separate commits, but I don't know if this is standard practice.

timj: it is standard practice.

r-owen: I'll just drop that change. I have a version of this file whose docs are cleaned up, but I don't want to mess up the merge of DM-13478 so I'll just hold on to it privately for now.

PR #17, opened 2018-06-01
DM-14620: Move ap_association DIA schemas their own module

morriscb:


python/lsst/ap/association/assoc_afw_utils.py
```diff
@@ -0,0 +1,279 @@
...
+import numpy as np
+
+import lsst.afw.table as afwTable
+
+__all__ = ["make_minimal_dia_object_schema",
+           "make_minimal_dia_source_schema",
+           "ccdVisitSchema"]
+
+
+"""Defines afw schemas and conversions for use in ap_association tasks.
```
parejkoj: module-level docstrings should be just under the copyright:

https://developer.lsst.io/python/numpydoc.html#placement-of-module-docstrings

python/lsst/ap/association/assoc_afw_utils.py
```diff
@@ -0,0 +1,279 @@
...
+
+Until a more finalized interface between the prompt products database (PPD) can
+be established, we put many utility functions for converting `lsst.afw.table` and
+`lsst.afw.image` objects to a PPD. This includes both mapping schemas between
+different catalogs and the DB.
+"""
+
+
+def make_minimal_dia_object_schema(filter_names=[]):
```
parejkoj: Don't default to an empty list (or any mutable). Use `None` instead, if it's optional, and set it to a list at the start of the method.

http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

python/lsst/ap/association/assoc_afw_utils.py
```diff
@@ -0,0 +1,279 @@
...
+        Minimal schema for DIAObjects.
+    """
+    schema = afwTable.SourceTable.makeMinimalSchema()
+    # For the MVP/S we currently only care about the position though
+    # in the future we will add summary computations for fluxes etc.
+    # as well as their errors.
+
+    # In the future we would like to store a covariance of the coordinate.
+    # This functionality is not defined in currently in the stack, so we will
+    # hold off until it is implemented. This is to be addressed in DM-7101.
```
parejkoj: Please mark this with a TODO: so it's easier to spot.

python/lsst/ap/association/assoc_afw_utils.py
```diff
@@ -0,0 +1,279 @@
...
+    schema : `lsst.afw.table.Schema`
+        Minimal schema for DIAObjects.
+    """
+    schema = afwTable.SourceTable.makeMinimalSchema()
+    # For the MVP/S we currently only care about the position though
+    # in the future we will add summary computations for fluxes etc.
+    # as well as their errors.
+
+    # In the future we would like to store a covariance of the coordinate.
+    # This functionality is not defined in currently in the stack, so we will
```
parejkoj: "not defined in currently in the stack"

python/lsst/ap/association/assoc_afw_utils.py
```diff
@@ -0,0 +1,279 @@
...
+def make_minimal_dia_source_schema():
+    """ Define and create the minimal schema required for a DIASource.
+
+    Return
+    ------
+    schema : `lsst.afw.table.Schema`
+        Minimal schema for DIAObjects.
+    """
+    schema = afwTable.SourceTable.makeMinimalSchema()
+    schema.addField("diaObjectId", type='L')
```
parejkoj: Please add `doc=' to document what all these fields are.

python/lsst/ap/association/assoc_afw_utils.py
```diff
@@ -0,0 +1,279 @@
...
+
+    # In the future we would like to store a covariance of the coordinate.
+    # This functionality is not defined in currently in the stack, so we will
+    # hold off until it is implemented. This is to be addressed in DM-7101.
+    schema.addField("pixelId", type='L')
+    schema.addField("nDiaSources", type='L')
+    for filter_name in filter_names:
+        schema.addField("psFluxMean_%s" % filter_name, type='D')
+        schema.addField("psFluxMeanErr_%s" % filter_name, type='D')
+        schema.addField("psFluxSigma_%s" % filter_name, type='D')
```
parejkoj: Please add `doc=` to document what all these fields are.

python/lsst/ap/association/assoc_afw_utils.py
```diff
@@ -0,0 +1,279 @@
...
+    schema.addField("ccdVisitId", type='L')
+    schema.addField("psFlux", type='D')
+    schema.addField("psFluxErr", type='D')
+    schema.addField("filterName", type='String', size=10)
+    schema.addField("filterId", type='L')
+    return schema
+
+
+def ccdVisitSchema():
+    """Define the schema for the CcdVisit table.
```
parejkoj: Document whether this is for an afw table, or for an sql table?

It should probably be called `getCcdVisitSchema()`, or `getCcdVisitSqlSchema()` if it's for SQL and not AFW.

python/lsst/ap/association/assoc_afw_utils.py
```diff
@@ -0,0 +1,279 @@
...
+        Exposure to store information from.
+
+    Returns
+    -------
+    values : `dict` of ``values``
+        List values representing info taken from the exposure.
+    """
+    visit_info = exposure.getInfo().getVisitInfo()
+    sphPoint = exposure.getWcs().getSkyOrigin()
+    flux0, flux0_err = exposure.getCalib().getFluxMag0()
```
parejkoj: I worry about storing the magnitude 0 point this way: PhotoCalibs will be more complicated, so this will only be an approximation.

python/lsst/ap/association/assoc_afw_utils.py
```diff
@@ -0,0 +1,279 @@
...
+    source_catalog : `lsst.afw.table.SourceCatalog`
+        Input source catalog to add alias columns to.
+    """
+    if not source_catalog.getSchema().contains(make_minimal_dia_source_schema()):
+        # Create aliases to appropriate flux fields if they exist.
+        schema_names = source_catalog.getSchema().getNames()
+        if 'base_PsfFlux_flux' in schema_names and \
+           'base_PsfFlux_fluxSigma' in schema_names and \
+           'psFlux' not in schema_names and \
+           'psFluxErr' not in schema_names:
```
parejkoj: I'm confused about the purpose of this `if` statement. When would the latter two fields not be in the schema, but the former two would be?

python/lsst/ap/association/assoc_afw_utils.py
```diff
@@ -0,0 +1,279 @@
...
+                                                      exposure=None):
+    """Create aliases in for the input source catalog schema and overwrite
+    values if requested.
+
+    Parameters
+    ----------
+    dia_sources : `lsst.afw.table.SourceCatalog`
+        Input source catalog to alias and overwrite values of.
+    obj_ids:  array-like of `int`
+        DIAObject ids to associated with the input DIASources. Overwrites
```
parejkoj: "to associate with"

Also, note that if not None, it should correspond 1:1 with `dia_sources`.

python/lsst/ap/association/assoc_afw_utils.py
```diff
@@ -0,0 +1,279 @@
...
+        if 'base_PsfFlux_flux' in schema_names and \
+           'base_PsfFlux_fluxSigma' in schema_names and \
+           'psFlux' not in schema_names and \
+           'psFluxErr' not in schema_names:
+                alias_map = source_catalog.getSchema().getAliasMap()
+                alias_map.set('psFlux', 'base_PsfFlux_flux')
+                alias_map.set('psFluxErr', 'base_PsfFlux_fluxSigma')
+
+
+def convert_dia_source_to_asssoc_schema_and_calibrate(dia_sources,
```
parejkoj: Why "and_calibrate"?

python/lsst/ap/association/assoc_afw_utils.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+def make_overwrite_dict(source_record, obj_id=None, exp_dict=None):
+    """Create a dictionary of values to overwrite the values stored in an
+    afw.table or when storing values in a DB.
+
+    Parameters
+    ----------
+    source_record : `lsst.afw.table.SourceRecord`
+        Input source catalog to modify values of if needed.
```
parejkoj: "if needed" implies that this is optional. Is it? What is the default value in that case?

python/lsst/ap/association/assoc_afw_utils.py
```diff
@@ -0,0 +1,279 @@
...
+        overwrite_dict['diaObjectId'] = int(obj_id)
+    if exp_dict is not None:
+        psFlux = source_record['psFlux']
+        psFluxErr = source_record['psFluxErr']
+
+        overwrite_dict['psFlux'] = psFlux / exp_dict['fluxZeroPoint']
+        overwrite_dict['psFluxErr'] = np.sqrt(
+            (psFluxErr / exp_dict['fluxZeroPoint']) ** 2 +
+            (psFlux * exp_dict['fluxZeroPointErr'] /
+             exp_dict['fluxZeroPoint'] ** 2) ** 2)
```
parejkoj: I don't like that we're coding the error calculation here explicitly. Call something from `Calib` or `PhotoCalib` instead. For example, `PhotoCalib.instFluxToMaggies(flux, fluxErr)` will give you the calibrated flux and fluxErr.

python/lsst/ap/association/assoc_afw_utils.py
```diff
@@ -0,0 +1,279 @@
+#
```
parejkoj: Why not call the file `afwUtils.py`? It's already in `ap.association`, so `ap.association.assoc_afw_utils` seems redundant. Also, files are camelCase:

https://developer.lsst.io/python/style.html#a-python-source-file-name-should-be-camelcase-with-leading-lowercase-and-ending-in-py

python/lsst/ap/association/assoc_db_sqlite.py
```diff
@@ -27,89 +27,25 @@
...
 import lsst.pex.config as pexConfig
-from lsst.pex.exceptions import InvalidParameterError
 import lsst.pipe.base as pipeBase
+from .assoc_afw_utils import \
+    make_minimal_dia_object_schema, \
+    make_minimal_dia_source_schema, \
+    ccdVisitSchema, \
+    add_dia_source_aliases_to_catalog, \
+    get_ccd_visit_info_from_exposure, \
+    make_overwrite_dict
```
parejkoj: Why not just `import .afwUtils` (per my above filename comment), and then they'll all be namespaced. You'll have to expand that file's `__all__`.

tests/test_association_db_sqlite.py
```diff
@@ -29,13 +29,14 @@
     AssociationDBSqliteTask, \
     AssociationDBSqliteConfig, \
     make_minimal_dia_source_schema
+from lsst.ap.association.assoc_afw_utils import \
```
parejkoj: Again, probably better to `from lsst.ap.association import afwUtils` and then you'll have the whole namespace.

python/lsst/ap/association/assoc_afw_utils.py
morriscb: I guess my response here is that there isn't really a better thing to store right now. The good news is that the database is storing the CccdVisitId which should allow us to look up the complicated calibration object regardless.

python/lsst/ap/association/assoc_afw_utils.py
morriscb: This is just to catch cases where someone may want to store a catalog that already has these aliases made or if I want to do a test case where I input a schema has psFlux etc.

python/lsst/ap/association/assoc_afw_utils.py
morriscb: ipDiffIm is currently putting a catalog that doesn't have any calibrated fluxes so I need to do that somewhere before storing the sources. I just called it "_and_calibrate" because I'm trying to be descriptive about what is fully happening in the function.

python/lsst/ap/association/assoc_afw_utils.py
morriscb: Do calexps or other exposure like objects have a PhotoCalib object? The calib object stored doesn't have an option to convert to maggies. Is there an example of how I can use a PhotoCalib object with an exposure and a catalog?

python/lsst/ap/association/assoc_afw_utils.py
parejkoj: Ah the joys of aliasing.

python/lsst/ap/association/assoc_afw_utils.py
```diff
@@ -0,0 +1,279 @@
...
+           'psFluxErr' not in schema_names:
+                alias_map = source_catalog.getSchema().getAliasMap()
+                alias_map.set('psFlux', 'base_PsfFlux_flux')
+                alias_map.set('psFluxErr', 'base_PsfFlux_fluxSigma')
+
+
+def convert_dia_source_to_asssoc_schema_and_calibrate(dia_sources,
+                                                      obj_ids=None,
+                                                      exposure=None):
+    """Create aliases in for the input source catalog schema and overwrite
```
parejkoj: typo here.

parejkoj: "and overwrite values if requested": what constitutes a request?

python/lsst/ap/association/assoc_afw_utils.py
parejkoj: "create_association_schema_from_dia_source" I think gets the point across more. You can say in the docstring that it calculates the calibrated on-sky fluxes; we may want to compute other ancillary information in the future, which is better put in the docstring.

python/lsst/ap/association/assoc_afw_utils.py
parejkoj: A more future-proof approach would be to create a `PhotoCalib` object when you read in the Exposure and extract the `Calib` and pass it around, instead of the zero-point. This should do it:

```
fluxMag0 = calib.getFluxMag0()
photoCalib = afwImage.PhotoCalib(1.0/fluxMag0[0], fluxMag0[1]/fluxMag0[0]**2, bbox)
```

That way, once we replace Calib with PhotoCalib (DM-10153), we just have to delete two lines.

Oh, please leave a TODO as well.

python/lsst/ap/association/assoc_db_sqlite.py
parejkoj: If you really don't want those things in the namespace, please make a note of that above the `__all__` in `afwUtils`, so that others don't expose it later.

python/lsst/ap/association/assoc_afw_utils.py
morriscb: Saying schema seems like I'm just making a schema. How about just "convert_dia_sources_to_asssoc_schema"?

parejkoj: That's fair. Though do you have to abbreviate `association` here?

PR #18, opened 2018-06-20
DM-14558: Load all DIASources at once in AssociationTask.update_dia_objects

morriscb:


python/lsst/ap/association/association.py
```diff
@@ -208,15 +211,15 @@ def update_dia_objects(self, dia_objects, updated_obj_ids, exposure):
                 updated_dia_objects.append(dia_object)
             dia_object.set('id', obj_id)
 
-            dia_sources = self.level1_db.load_dia_sources([obj_id])
+            obj_dia_sources = dia_sources[associated_ids == obj_id].copy(deep=True)
```
mrawls: The variable name `obj_dia_sources` is not very clear. I think you're retrieving (copying) all of the DIASources associated with an object ID. But isn't `associated_ids` a whole bunch of object IDs while `obj_id` is a single object ID inside this for-loop? I'm missing something about how this works.

mrawls: Nevermind... this was apparently an outdated version of what you pushed... weird GitHub is weird.

python/lsst/ap/association/association.py
```diff
@@ -208,15 +212,17 @@ def update_dia_objects(self, dia_objects, updated_obj_ids, exposure):
                 updated_dia_objects.append(dia_object)
             dia_object.set('id', obj_id)
 
-            dia_sources = self.level1_db.load_dia_sources([obj_id])
+            start_idx = dia_sources.lower_bound(obj_id, diaObjectId_key)
+            end_idx = dia_sources.upper_bound(obj_id, diaObjectId_key)
+            obj_dia_sources = dia_sources[start_idx:end_idx].copy(deep=True)
```
mrawls: Everything here makes sense now, but I'm not a fan of this variable name. It's a copy of the DIASources that correspond to a single objectID, correct? Maybe just add an in-line comment or two to make it easier to follow if you can't think of something shorter than "dia_sources_corresponding_to_obj_id" 😄

PR #20, opened 2018-06-25
DM-14885: Use either nanoseconds or MJD in PPDB CcdVisit table

morriscb:


python/lsst/ap/association/afwUtils.py
```diff
@@ -163,7 +164,7 @@ def get_ccd_visit_info_from_exposure(exposure):
               'ra': sphPoint.getRa().asDegrees(),
               'decl': sphPoint.getDec().asDegrees(),
               'expTime': visit_info.getExposureTime(),
-              'expMidptMJD': visit_info.getDate().nsecs() * 10 ** -9,
+              'expMidptMJD': date.get(date.MJD),
```
mrawls: Perhaps have it as `date.get(system=DateTime.MJD)` to be more explicit? (This is what Eli says he uses on Slack in the dm channel)

Please also update the values list comment above; the field name is `expMidptMJD` but it is described above as `dateTimeMJD` in seconds. Otherwise 💯 and many thanks

PR #27, opened 2018-09-26
DM-15775: Migrate to dax_ppdb instead of l1dbproto

andy-slac:
l1dbproto package was replaced by dax_ppdb, need to rename imports and
class names here, and update table file. All unit tests pass after the
renaming.


python/lsst/ap/association/l1dbproto.py
```diff
@@ -146,7 +147,7 @@ def load_dia_objects(self, exposure):
 
         indexer_indices, on_boundry = self.indexer.getShardIds(
             ctr_coord, max_radius)
-        # Index types must be cast to int to work with l1dbproto.
+        # Index types must be cast to int to work with dax.ppdb.
```
kfindeisen: I think `dax_ppdb` would be more understandable.

PR #31, opened 2018-10-29
DM-16299: Create placeholder AP DPDDifying task

morriscb:


python/lsst/ap/association/apDataProductMaker.py
```diff
@@ -0,0 +1,196 @@
...
+        inputCatalog: `lsst.afw.table.SourceCatalog`
+           Input catalog with data to be copied into new output catalog.
+
+        Returns
+        -------
+        outputCatalog: `lsst.afw.table.SourceCatalog`
+            Output catalog with data copied from input and new column names.
+        """
+        outputCatalog = afwTable.SourceCatalog(self.outputSchema)
+        outputCatalog.preallocate(len(inputCatalog))
```
yalsayyad: This doesn't look like the right way to do this.  Have you seen:
https://github.com/lsst-dm/Oct15_bootcamp/blob/measurement/measurement/measurement.pdf?raw=true

Jim says: `extend` is more efficient than `preallocate` + `for` loop.  (Also, in cases where the `for` loop is necessary, `reserve` is better than `preallocate`)

python/lsst/ap/association/apDataProductMaker.py
```diff
@@ -0,0 +1,196 @@
+#
```
yalsayyad: Did you forget to rename the module? Typically we name the files something similar to the Task it contains so that it's easy to find. I wouldn't necessarily know that `apDataProductMaker.py` contains `APDataMapperTask`. We also usually name our tasks with the verb first: so `MapAPDataTask` and the module `mapApData.py`

python/lsst/ap/association/apDataProductMaker.py
```diff
@@ -0,0 +1,196 @@
...
+            Output catalog with data copied from input and new column names.
+        """
+        outputCatalog = afwTable.SourceCatalog(self.outputSchema)
+        outputCatalog.preallocate(len(inputCatalog))
+
+        for inputRecord in inputCatalog:
+            outputRecord = outputCatalog.addNew()
+            outputRecord.assign(inputRecord, self.mapper)
+
+        if outputCatalog.isContiguous():
```
yalsayyad: In your testing, have you found that this is needed?   

Jim says: "I also think the `isContiguous` check should be unnecessary; maybe best to turn that into an `assert` to make sure I'm not wrong about that and document-in-code that it's guaranteed."

Reviewer's personal preference is throwing exceptions instead of asserts, but coder's choice.

tests/test_ap_data_product_maker.py
```diff
@@ -0,0 +1,234 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import numpy as np
+import unittest
+
+from lsst.ap.association import \
```
yalsayyad: We discourage `\` over 
```
from lsst.ap.association import (APDataMapperConfig,
    APDataMapperTask, 
    DiaSourceMapperConfig, 
    DiaSourceMapperTask)
```
https://developer.lsst.io/python/style.html#python-s-implied-continuation-inside-parens-brackets-and-braces-should-be-used-for-wrapped-lines

tests/test_ap_data_product_maker.py
```diff
@@ -0,0 +1,234 @@
+# This file is part of ap_association.
```
yalsayyad: Hurray! Unit test! 🥇

tests/test_ap_data_product_maker.py
```diff
@@ -0,0 +1,234 @@
...
+        for inObj, outObj in zip(self.inputCatalog, outputCatalog):
+            self.assertEqual(
+                outObj["ccdVisitId"],
+                self.exposure.getInfo().getVisitInfo().getExposureId())
+            self.assertEqual(
+                outObj["midPointTai"],
+                self.exposure.getInfo().getVisitInfo().getDate().get(
+                    system=dafBase.DateTime.MJD))
+            for inputName, outputName in apDMapConfig.copyColumns.items():
+                if inputName[:4] == "slot":
```
yalsayyad: Also see python's `inputName.startswith("slot")`

python/lsst/ap/association/apDataProductMaker.py
```diff
@@ -0,0 +1,196 @@
...
+        """
+        visit_info = exposure.getInfo().getVisitInfo()
+        ccdVisitId = visit_info.getExposureId()
+        midPointTaiMJD = visit_info.getDate().get(system=DateTime.MJD)
+
+        flux0, flux0Err = exposure.getCalib().getFluxMag0()
+        photoCalib = afwImage.PhotoCalib(1 / flux0, flux0Err / flux0 ** 2)
+
+        outputCatalog = afwTable.SourceCatalog(self.outputSchema)
+        outputCatalog.preallocate(len(inputCatalog))
```
yalsayyad: see `reserve` suggestion above.

python/lsst/ap/association/apDataProductMaker.py
morriscb: Moved the first Task to the extend method and the DiaSource task to reserve. Thanks for sending this. Finding examples of how to use mappers within the python side of the stack is rather difficult.

jdswinbank: I learnt things!

(Not least that finding good documentation on afw::table is hard... :))

PR #32, opened 2018-11-19
DM-15588: Remove home-brewed SQLite PPDB

morriscb:


tests/test_association_task.py
```diff
@@ -141,16 +232,18 @@ def test_run(self):
...
                 # Test that associating to the existing DIAObjects went
                 # as planned and test that the IDs of the newly associated
                 # DIASources is correct.
-                self.assertEqual(dia_object['nDiaSources'], 3)
+                self.assertEqual(dia_object['gPSFluxNdata'], 2)
+                self.assertEqual(dia_object['rPSFluxNdata'], 1)
                 self.assertEqual(dia_object.getId(), obj_idx)
             else:
-                self.assertEqual(dia_object['nDiaSources'], 1)
+                self.assertEqual(dia_object['gPSFluxNdata'], 1)
```
kfindeisen: Why not use `nDiaSources`? You do set it...

tests/test_association_task.py
```diff
@@ -110,9 +190,20 @@ def setUp(self):
...
-            bbox=bbox,
-            wcs=wcs,)
+
+        ctr_coord = wcs.pixelToSky(bbox.getCenter())
+        max_radius = max(
+            ctr_coord.separation(wcs.pixelToSky(pp))
+            for pp in bbox.getCorners())
+
+        assoc_config = AssociationConfig()
+        assoc_task = AssociationTask(config=assoc_config)
```
kfindeisen: This is equivalent to
```suggestion
        assoc_task = AssociationTask()
```

tests/test_association_task.py
```diff
@@ -192,24 +285,33 @@ def _run_association_and_retrieve_objects(self, create_objects=False):
...
+            dia_source["psFluxErr"] = \
+                np.sqrt((100 / self.flux0) ** 2 + (10000 * self.flux0_err / self.flux0 ** 2) ** 2)
+            dia_source["apFlux"] = 10000 / self.flux0
+            dia_source["apFluxErr"] = \
+                np.sqrt((100 / self.flux0) ** 2 + (10000 * self.flux0_err / self.flux0 ** 2) ** 2)
+            dia_source["filterName"] = self.exposure.getFilter().getName()
+            dia_source["filterId"] = self.exposure.getFilter().getId()
+            dia_source["x"] = 0
+            dia_source["y"] = 0
+            dia_source["snr"] = 10
```
kfindeisen: Are these all mandatory columns? :fearful: I'm a bit worried about code duplication with the code you're supposed to be testing...

tests/test_association_task.py
```diff
@@ -240,12 +341,15 @@ def _store_dia_objects_and_sources(self):
...
-                dia_object['psFluxMeanErr_%s' % filter_name] = 1
-                dia_object['psFluxSigma_%s' % filter_name] = 1
-        assoc_db.store_dia_objects(dia_objects, True)
+                dia_object["pixelId"] = 225823
+                dia_object['%sPSFluxMean' % filter_name] = 1
+                dia_object['%sPSFluxMeanErr' % filter_name] = 1
+                dia_object['%sPSFluxSigma' % filter_name] = 1
+                dia_object['%sPSFluxNdata' % filter_name] = 1
+
+        dateTime = dafBase.DateTime(nsecs=1400000000 * 10**9 - 1000)
```
kfindeisen:
```suggestion
        dateTime = dafBase.DateTime(nsecs=1400000000 * 1e9 - 1000)
```
This is kind of unusual notation anyway; why not just write out the number?

tests/test_association_task.py
```diff
@@ -258,20 +362,26 @@ def _store_dia_objects_and_sources(self):
...
             dia_source["psFlux"] = 10000 / self.flux0
-            dia_source["psFluxErr"] = np.sqrt(
-                (100 / self.flux0) ** 2 +
-                (10000 * self.flux0_err / self.flux0 ** 2) ** 2)
+            dia_source["psFluxErr"] = \
+                np.sqrt((100 / self.flux0) ** 2 + (10000 * self.flux0_err / self.flux0 ** 2) ** 2)
+            dia_source["apFlux"] = 10000 / self.flux0
+            dia_source["apFluxErr"] = \
+                np.sqrt((100 / self.flux0) ** 2 + (10000 * self.flux0_err / self.flux0 ** 2) ** 2)
+            dia_source["snr"] = 10
```
kfindeisen: Some code duplication with `_run_association_and_retrieve_objects` and `test_update_dia_objects`; can you combine how the three methods initialize sources?

tests/test_association_task.py
```diff
@@ -286,66 +396,68 @@ def test_update_dia_objects(self):
...
 
         # Store them in the DB containing pre-existing objects.
-        assoc_db_config = AssociationDBSqliteConfig()
-        assoc_db_config.db_name = self.db_file
-        assoc_db_config.filter_names = self.filter_names
-        assoc_db = AssociationDBSqliteTask(config=assoc_db_config)
-        assoc_db.create_tables()
+        ppdb = Ppdb(config=self.ppdbConfig,
+                    afw_schemas=dict(DiaObject=make_dia_object_schema(),
+                                     DiaSource=make_dia_source_schema()))
```
kfindeisen: Code duplication with `_store_dia_objects_and_sources`; maybe have a `makePpdb` function?

tests/test_association_task.py
```diff
@@ -286,66 +396,68 @@ def test_update_dia_objects(self):
...
-        output_dia_objects = assoc_task.level1_db.load_dia_objects(self.exposure)
+        output_dia_objects = ppdb.getDiaObjects(self.index_ranges)
 
         # Data and column names to test.
         test_column_names = [
-            'id', 'nDiaSources',
-            'psFluxMean_g', 'psFluxMeanErr_g', 'psFluxSigma_g',
-            'psFluxMean_r', 'psFluxMeanErr_r', 'psFluxSigma_r']
+            'id', 'gPSFluxMean', 'gPSFluxMeanErr', 'gPSFluxSigma',
+            'rPSFluxMean', 'rPSFluxMeanErr', 'rPSFluxSigma']
```
kfindeisen: Does this depend on the content of the schema files or some other external source? I'm worried about it getting out of sync.

kfindeisen: Actually, why not just use `test_dia_object_values[0].keys()`? You'd have one fewer copy to keep up to date.

python/lsst/ap/association/afwUtils.py
```diff
@@ -59,54 +59,453 @@ def make_minimal_dia_object_schema(filter_names=None):
...
-        schema.addField("psFluxMean_%s" % filter_name, type='D',
-                        doc='Calibrated mean flux in %s band.' % filter_name)
-        schema.addField("psFluxMeanErr_%s" % filter_name, type='D',
-                        doc='Calibrated error on mean flux in %s band.' %
-                        filter_name)
-        schema.addField("psFluxSigma_%s" % filter_name, type='D',
-                        doc='Calibrated scatter in flux in %s band.' %
-                        filter_name)
+
+    # Generated automatically from ppdb-schema.yaml in dax_ppdb/data.
```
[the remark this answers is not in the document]
morriscb: My understanding is that the underlying cat schema format (which the Ppdb schema is auto generated from) is changing in the near future. I didn't want completely auto generate the afw schema until the new cat schema format is finalized.

python/lsst/ap/association/afwUtils.py
```diff
@@ -59,54 +59,453 @@ def make_minimal_dia_object_schema(filter_names=None):
...
+    schema.addField('yFPFluxMean', type='D',
+                    doc='Weighted mean forced photometry flux for y filter.')
+    schema.addField('yFPFluxMeanErr', type='D',
+                    doc='Standard error of yFPFluxMean.')
+    schema.addField('yFPFluxSigma', type='D',
+                    doc='Standard deviation of the distribution of yFPFlux.')
+    # Mapping of arrays and BLOBs is not currently supported by the PPDB so we
+    # these columns out.
+    """
+    schema.addField('uLcPeriodic', type='ArrayF',
```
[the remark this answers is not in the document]
morriscb: Same. I'll change these to normal comments.

python/lsst/ap/association/association.py
```diff
@@ -130,22 +143,91 @@ def run(self, dia_sources, exposure):
             Input exposure representing the region of the sky the dia_sources
             were detected on. Should contain both the solved WCS and a bounding
             box of the ccd.
+        ppdb : `lsst.dax.ppdb.Ppdb`
+            Ppdb connection object to retrieve DIASources/Objects from and
+            write to.
```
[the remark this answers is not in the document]
morriscb: Seems so? ap_association doesn't set the db location anymore but instead receives a Ppdb object. The settings for the Ppdb are a PexConfig object though and currently ap_verify and ap_pipe set these configs. Not sure if that completely answers the question.

python/lsst/ap/association/afwUtils.py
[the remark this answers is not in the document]
kfindeisen: Ok, but what do you mean by the comment? Is `afwUtils.py` being auto-generated somehow?

tests/test_association_task.py
```diff
@@ -192,24 +285,33 @@ def _run_association_and_retrieve_objects(self, create_objects=False):
...
-        assoc_config.level1_db.value.db_name = self.db_file
-        assoc_config.level1_db.value.filter_names = self.filter_names
         assoc_task = AssociationTask(config=assoc_config)
 
-        assoc_task.run(dia_sources, self.exposure)
-        assoc_task.level1_db.close()
+        ppdb = Ppdb(config=self.ppdbConfig,
+                    afw_schemas=dict(DiaObject=make_dia_object_schema(),
+                                     DiaSource=make_dia_source_schema()))
+        ppdb._schema.makeSchema()
```
kfindeisen: Given the discussion on lsst/dax_ppdb#5, this line should be `ppdb.makeSchema()` but otherwise doesn't need to be changed.

tests/test_association_task.py
```diff
@@ -221,11 +323,10 @@ def _store_dia_objects_and_sources(self):
...
         # Create an empty database
-        assoc_db_config = AssociationDBSqliteConfig()
-        assoc_db_config.db_name = self.db_file
-        assoc_db_config.filter_names = self.filter_names
-        assoc_db = AssociationDBSqliteTask(config=assoc_db_config)
-        assoc_db.create_tables()
+        ppdb = Ppdb(config=self.ppdbConfig,
+                    afw_schemas=dict(DiaObject=make_dia_object_schema(),
+                                     DiaSource=make_dia_source_schema()))
+        ppdb._schema.makeSchema()
```
kfindeisen: Given the discussion on lsst/dax_ppdb#5, this line should be `ppdb.makeSchema()` but otherwise doesn't need to be changed (or it could be merged with the similar code in the test case just above it).

data/ppdb-ap-pipe-schema-extra.yaml
```diff
@@ -13,6 +13,10 @@ columns:
   nullable: true
   description: Time when validity of this diaObject ends.
   default: null
+- name: nDiaSources
+  type: INT
+  nullable: false
+  description: Total number of DiaSources associated with this DiaObject.
```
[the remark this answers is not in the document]
morriscb: The one being used in ap_pipe is from ap_association. The reason the duplication exists is because the ap_association one as extra columns for the observed DiaSource filter.  It also doesn't have the DiaObjectLast table as we don't currently use the Ppdb concept of "dailyJob".

tests/test_association_task.py
kfindeisen: Sounds good.

tests/test_association_task.py
morriscb: Does the new make_schema function satisfy this?

tests/test_association_task.py
morriscb: Ditto with above.

tests/test_association_task.py
kfindeisen: If you really meant `_make_ppdb`, then mostly yes. Did you mean to leave off the `drop=True` bit?

tests/test_association_task.py
morriscb: Changed to a string input.

python/lsst/ap/association/afwUtils.py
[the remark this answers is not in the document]
morriscb: It was a quick and dirty solution to comment out the block because Ppdb doesn't currently support array types or BLOBs. Forgot to change it for the PR.

python/lsst/ap/association/afwUtils.py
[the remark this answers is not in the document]
morriscb: The full file isn't, but the afw columns are.

tests/test_association_task.py
morriscb: "x" and "y" are non-Nullable in the Ppdb thus I set them here. The association step takes place in RA/DEC and these columns are not touched.

PR #36, opened 2018-12-14
DM-16650: update to new nJy PhotoCalib definition

parejkoj:


tests/test_map_ap_data.py
```diff
@@ -110,9 +110,14 @@ def setUp(self):
...
-        self.flux0 = 2
-        self.flux0Err = 1
-        self.exposure.getCalib().setFluxMag0((self.flux0, self.flux0Err))
+        # TODO: Need to save the PhotoCalib until DM-10153 is done and Calib is replaced by PhotoCalib.
+        scale = 2
+        scaleErr = 1
+        self.photoCalib = afwImage.PhotoCalib(scale, scaleErr)
+        flux0 = self.photoCalib.getInstFluxAt0Magnitude()
+        referenceFlux = 1e23 * 10**(48.6 / -2.5) * 1e9
+        flux0Err = (scaleErr / scale**2 * referenceFlux)
```
erykoff: The order-of-operations would be clearer if `referenceFlux` is at the front.

PR #38, opened 2019-01-29
DM-14685: Replace afw.geom with lsst.geom in ap_association

morriscb:


python/lsst/ap/association/association.py
```diff
@@ -54,7 +54,7 @@ def _set_mean_position(dia_object_record, dia_sources):
         Average position of the dia_sources.
```
kfindeisen: Type `lsst.afw.geom.SpherePoint` needs to be updated.

kfindeisen: There's also a reference to `lsst.afw.geom.Angle` in the same file, but it's too far from any changes for me to flag it.

morriscb: Thanks for catching these. Realized I just searched for afwGeom the alias not also the full name. I'll double check and update.

PR #40, opened 2019-02-12
DM-15795: Create DiaForcedSourceTask

morriscb:


python/lsst/ap/association/diaForcedSource.py
```diff
@@ -0,0 +1,170 @@
...
+              image at the DiaObject location (`lsst.afw.table.SourceCatalog`)
+            - ``directForcedSources`` : Psf forced photometry on the direct
+              image at the DiaObject location (`lsst.afw.table.SourceCatalog`)
+        """
+        idFactoryDiff = afwTable.IdFactory.makeSource(
+            diffim.getInfo().getVisitInfo().getExposureId(),
+            64 - expIdBits)
+        idFactoryDirect = afwTable.IdFactory.makeSource(
+            diffim.getInfo().getVisitInfo().getExposureId(),
+            64 - expIdBits)
```
isullivan: Could you use something like DM-17689 to avoid hard-coded numbers here?

tests/test_diaForcedSource.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+def create_test_dia_objects(n_points, wcs, startPos=100):
+    """Create dummy DIASources or DIAObjects for use in our tests.
+    Parameters
+    ----------
+    n_points : `int`
+        Number of DiaObject test points to create.
+    wcs : `lsst.afw.geom.SkyWcs`
+        Wcs to convert RA/Dec to x/y if provided.
```
isullivan: What do you mean by "if provided" here? It doesn't look optional.

tests/test_diaForcedSource.py
```diff
@@ -0,0 +1,207 @@
...
+        self.flux0_err = 100
+        self.exposureId = 1234
+        self.exposureTime = 200.
+        self.imageSize = [1024, 1153]
+        self.dateTime = "2014-05-13T17:00:00.000000000"
+
+        # Make images  with one source in them and distinct values and
+        # variance for each image.
+        # Direct Image
+        ms = afwImage.MaskedImageF(lsst.geom.ExtentI(self.imageSize[0] + 1,
```
isullivan: "ms" is not a very descriptive variable name. Is the "m" for masked image and the "s" for source? I'd prefer to see a slightly longer name, but I guess this is ok for a test.

tests/test_diaForcedSource.py
```diff
@@ -0,0 +1,207 @@
...
+        # Difference Image
+        ms = afwImage.MaskedImageF(lsst.geom.ExtentI(self.imageSize[0] + 1,
+                                                     self.imageSize[1] + 1))
+        ms.image[100, 100, afwImage.LOCAL] = 20
+        ms.getVariance().set(2)
+        bbox = lsst.geom.BoxI(
+            lsst.geom.PointI(1, 1),
+            lsst.geom.ExtentI(self.imageSize[0],
+                              self.imageSize[1]))
+        self.mi = afwImage.MaskedImageF(ms, bbox, afwImage.LOCAL)
```
isullivan: Does `self.mi` need to be saved as a class attribute? You're just overwriting the value you set it to a few lines back, which suggests you don't need it outside of setting up `self.exposure` and `self.diffim`.

tests/test_diaForcedSource.py
```diff
@@ -0,0 +1,207 @@
...
+        self.diffim.setDetector(detector)
+        self.diffim.getInfo().setVisitInfo(visit)
+        self.diffim.setFilter(afwImage.Filter('g'))
+        self.diffim.getCalib().setFluxMag0((self.flux0, self.flux0_err))
+
+        self.expIdBits = 16
+
+        FWHM = 5
+        psf = measAlg.DoubleGaussianPsf(
+            15, 15, FWHM/(2*np.sqrt(2*np.log(2))))
```
isullivan: I think this could fit on one line.

tests/test_diaForcedSource.py
morriscb: Removed. I re-purposed a previous method.

tests/test_diaForcedSource.py
morriscb: sorry, these were copy-pasted from a different unittest. I'll give them slightly more descriptive names.

PR #42, opened 2019-03-05
DM-18196: Fix afw schema missmatch between ap_association and dax_ppdb

morriscb:


python/lsst/ap/association/afwUtils.py
```diff
@@ -230,29 +234,31 @@ def make_dia_object_schema(filter_names=None):
...
-    schema.addField('nearbyObj3LnP', type='D',
+    schema.addField('nearbyObj3LnP', type='F',
                     doc='Natural log of the probability that the observed diaObject is the same as the '
                         'nearbyObj3.')
     schema.addField('flags', type='L',
                     doc='Flags, bitwise OR tbd.')
     schema.addField('pixelId', type='L',
                     doc='HTM index.')
+    schema.addField('lastNonForcedSource', type='L',
+                    doc='Last time when non-forced DIASource was seen for this object')
```
andy-slac: I'm not sure that you want this column on your AFW schema - this is an implementation detail internal to dax_ppdb and you are not supposed to depend on it, I think this schema should only depend on things that are specified in PPDB. I guess you want AFW schema returned from database to be exactly the same as what you saved? If you cannot tolerate few extra columns in returned AFW schema then we should filter out those internal columns from returned result.

And I think I'm OK with this change for now if that helps you fixing issues but you should be aware that dax_ppdb internals are not stable and be prepared to update this schema if those internals change.

tests/test_association_task.py
```diff
@@ -390,6 +390,8 @@ def _store_dia_objects_and_sources(self):
         dateTime = dafBase.DateTime("2014-05-13T16:00:00.000000000",
                                     dafBase.DateTime.Timescale.TAI)
         ppdb.storeDiaObjects(dia_objects, dateTime.toPython())
+        loaded_dia_objects = ppdb.getDiaObjects(self.index_ranges)
+        self.assertTrue(loaded_dia_objects.schema == dia_objects.schema)
```
andy-slac: I do not think that dax_pddb provides (or should provide) this sort of guarantee today, but maybe I misunderstand the meaning of this test. Do you want to test that catalog that you _save_ has a complete schema?

morriscb: The reason the schemas need to match is that later down the road I use the afw catalog of DiaObjects in a measurement task. The schema of the DiaObjects must match the schema that I used to initialize the task. So we can either have the schema match up on output for I would have to do an intermediate translation to a different catalog and then input into the task.

Either way, and related to the above comment, it would likely be helpful to have a few things down the road for the ppdb:

1. Given a ppdb config, a method to return what the afw or similar data object schema will be on call of getDiaObjects and getDiaSources. This would remove a bit of code from ap_association and make it harder for the schemas in the two modules to get our of sync.
2. Hiding the internal ppdb variables like `lastNonForcedSource` from the returns. What happens currently is that if the ppdb config for return columns of the DiaObjects is set to `[]`, the internal columns are grabbed as well. It would be great to have them always hidden unless specifically requested in the output columns.

Anywho, those are my two main ideas going forward, anything there sound worth ticketing or should we just wait for further versions of the ppdb or the ScienceDataModel converter?

python/lsst/ap/association/afwUtils.py
morriscb: See the below comment. I don't really need these columns to exist, but as is can be returned by the ppdb so to make the schemas match I include it here.

tests/test_association_task.py
andy-slac: Agree with both points. Please open a ticket for me so that at least I don't forget about it :)

PR #43, opened 2019-03-15
DM-18318: Create initial subset of timeseries features for DIAObject

morriscb:


data/ppdb-ap-pipe-schema-extra.yaml
```diff
@@ -18,6 +18,419 @@ columns:
...
+  unit: nJy
+- name: uPSFluxStetsonJ
+  type: FLOAT
+  nullable: true
+  description: StetsonJ statistic for the u band fluxes.
+- name: uPSFluxLinearSlope
+  type: FLOAT
+  nullable: true
+  description: Linear best fit slope of the u band fluxes.
+  unit: nJy/MJD
```
ebellm: should be `nJy/day` (MJD is a time system, not a time unit).  Same for the other slopes here.

python/lsst/ap/association/afwUtils.py
```diff
@@ -104,6 +104,7 @@ def make_dia_object_schema(filter_names=None):
                     doc='Standard error of uPSFluxMean.')
     schema.addField('uPSFluxSigma', type='F',
                     doc='Standard deviation of the distribution of uPSFlux.')
+
```
ebellm: extraneous whitespace?

python/lsst/ap/association/afwUtils.py
```diff
@@ -120,6 +121,7 @@ def make_dia_object_schema(filter_names=None):
                     doc='Standard error of gPSFluxMean.')
     schema.addField('gPSFluxSigma', type='F',
                     doc='Standard deviation of the distribution of gPSFlux.')
+
```
ebellm: extraneous whitespace?

python/lsst/ap/association/association.py
```diff
@@ -75,27 +78,173 @@ def _set_flux_stats(dia_object_record, dia_sources, filter_name, filter_id):
...
+        Slope (m) and intercept (b) values fit to the light-curve.
+    """
+    def model(x):
+        return ((x[0] * times + x[1] - fluxes) / errors) ** 2
+
+    ans = least_squares(model, x0=[0., np.mean(fluxes)]).x
+    return ans
+
+
+def _stentson_J(fluxes, errors):
```
ebellm: "stentson" -> "stetson" (someone's name); fix throughout

python/lsst/ap/association/association.py
```diff
@@ -75,27 +78,173 @@ def _set_flux_stats(dia_object_record, dia_sources, filter_name, filter_id):
...
+
+    totFluxes = dia_sources.get("totFlux")[currentFluxMask]
+    totFluxErrors = dia_sources.get("totFluxErr")[currentFluxMask]
+    noNanMask = np.logical_and(np.isfinite(totFluxes),
+                               np.isfinite(totFluxErrors))
+    totFluxes = totFluxes[noNanMask]
+    totFluxErrors = totFluxErrors[noNanMask]
+
+    if len(totFluxes) == 1:
+        dia_object_record['%sFPFluxMean' % filter_name] = totFluxes
```
ebellm: Let's call these `totFluxMean` and so on so the link to the `DIASource` quantity is more clear

python/lsst/ap/association/association.py
```diff
@@ -75,27 +78,173 @@ def _set_flux_stats(dia_object_record, dia_sources, filter_name, filter_id):
...
+    p_k = delta_val ** 2 - 1
+
+    return np.mean(np.sign(p_k) * np.sqrt(np.fabs(p_k)))
+
+
+def _stentson_mean(values, errors, alpha=2., beta=2., n_iter=20, tol=1e-6):
+    """Compute the Stentson mean of the fluxes which down-weights outliers.
+
+    Weighted biased on an error weighted difference scaled by a constant
+    (1/``a``) and raised to the power beta. Higher beta's more harshly penalize
```
ebellm: "beta's" -> "betas" (or "values of beta")

python/lsst/ap/association/association.py
```diff
@@ -75,27 +78,173 @@ def _set_flux_stats(dia_object_record, dia_sources, filter_name, filter_id):
...
+    ----------
+    fluxes : `numpy.ndarray` (N,)
+        Calibrated lightcurve flux values.
+    errors : `numpy.ndarray`
+        Errors on the calibrated lightcurve fluxes.
+
+    Returns
+    -------
+    stentsonJ : `float`
+        StentsonJ statistic for the input fluxes and errors.
```
ebellm: I'd add a citation note to Stetson 1996

python/lsst/ap/association/association.py
```diff
@@ -75,27 +78,173 @@ def _set_flux_stats(dia_object_record, dia_sources, filter_name, filter_id):
...
+    n_iter : `int`
+        Number of iterations of clipping.
+    tol : `float`
+        Fractional and absolute tolerance goal on the change in the mean before
+        exiting early.
+
+    Returns
+    -------
+    wmean : `float`
+        Weighted Stentson mean result.
```
ebellm: I'd add a citation note to Stetson 1996

python/lsst/ap/association/association.py
morriscb: That's what I initially had too, however, FPFlux is coming from the current version of the cat schema and the schema in dax_ppdb. I'm trying to reuse those two as much as possible so things don't get even more confused with the column names. I can add totFlux columns to the schema extra files and just leave the FPF names empty but that might be confusing for folks using the current setup. Hopefully once we have a automated way of making yaml schemas from the DPDD soon.

ebellm: I just tend to think FPFlux should refer to a different thing, which is means/errors on the diffim forced fluxes.

morriscb: Okay, so would you prefer that I add them in as specifically totFlux here and in the schema files?

ebellm: Yes, that would be my preference

PR #45, opened 2019-04-02
DM-18189: Investigate mapping of ip_diffim dipole fields to DPDD

morriscb:


python/lsst/ap/association/mapApData.py
```diff
@@ -230,16 +264,30 @@ def run(self, inputCatalog, exposure):
 
         return outputCatalog
 
+    def computeSnr(self, inputRecord, outputRecord):
```
ebellm: Since the DPDD wants this to be detection SNR specifically, using the measurements in existing catalogs this way is misleading.  I'd suggest simply removing this (and related SNR pieces) for now.

ebellm: Since the DPDD wants this to be detection SNR specifically, using the measurements in existing catalogs this way is misleading. I'd suggest simply removing this (and related SNR pieces) for now.

python/lsst/ap/association/mapApData.py
```diff
@@ -134,6 +146,24 @@ class MapDiaSourceConfig(pexConfig.Config):
...
+            "positive flux lobes.",
+        default="ip_diffim_DipoleFit",
+    )
+    dipSepColumn = pexConfig.Field(
+        dtype=str,
+        doc="Column of the separation of the negative and positive poles of "
+            "the dipole.",
+        default="ip_diffim_DipoleFit_separation"
+    )
+    snrFlux = pexConfig.Field(
```
ebellm: (as below, let's strike this for now)

python/lsst/ap/association/mapApData.py
```diff
@@ -217,7 +248,10 @@ def run(self, inputCatalog, exposure):
         for inputRecord in inputCatalog:
             outputRecord = outputCatalog.addNew()
             outputRecord.assign(inputRecord, self.mapper)
+            self.computeSnr(inputRecord, outputRecord)
```
ebellm: (strike)

python/lsst/ap/association/mapApData.py
```diff
@@ -250,15 +298,61 @@ def calibrateFluxes(self, inputRecord, outputRecord, photoCalib):
...
+            0.5 * np.sqrt(neg_meas.error ** 2 + pos_meas.error ** 2))
+        outputRecord.set(
+            "dipFluxDiff",
+            np.abs(pos_meas.value) - np.abs(neg_meas.value))
+        outputRecord.set(
+            "dipFluxDiffErr",
+            np.sqrt(neg_meas.error ** 2 + pos_meas.error ** 2))
+
+    def computeDipoleSep(self, inputRecord, outputRecord, wcs):
+        """Calibrate and compute dipole mean flux and diff flux.
```
ebellm: Shouldn't this docstring be "compute dipole separation"?

python/lsst/ap/association/mapApData.py
morriscb: What about using PeakFlux? Would that be an acceptable middle ground while not perfectly being the `detection` snr?

python/lsst/ap/association/mapApData.py
morriscb: Thanks, for catching this. I copy pasted the previous method's docs and forgot to change the description.

python/lsst/ap/association/mapApData.py
morriscb: As we discussed in the `ap_pipe`, I'll remove this to avoid any confusion between the what the value here means the one requested by the Ppdb.

PR #50, opened 2019-08-01
DM-20760: Drop DiaSources if RA/DEC is NaN and throw warning.

morriscb:


python/lsst/ap/association/association.py
```diff
@@ -405,6 +406,36 @@ def _check_dia_object_position(self, dia_object_record, bbox, wcs):
...
+        dia_sources : `pandas.DataFrame`
+            Input DiaSources to check for NaN values.
+
+        Returns
+        -------
+        trimmed_sources : `pandas.DataFrame`
+            DataFrame of DiaSources trimmed of all entries with NaN values for
+            RA/DEC.
+        """
+        nan_mask = np.logical_or(np.isnan(dia_sources.loc[:, "ra"]),
```
ebellm: I would do this in a more native pandas way:

`nan_mask = dia_sources["ra"].isnull() |  dia_sources["decl"].isnull()`

python/lsst/ap/association/association.py
```diff
@@ -405,6 +406,36 @@ def _check_dia_object_position(self, dia_object_record, bbox, wcs):
...
+                                 np.isnan(dia_sources.loc[:, "decl"]))
+        if np.any(nan_mask):
+            nan_idxs = np.argwhere(nan_mask).flatten()
+            for nan_idx in nan_idxs:
+                self.log.warning(
+                    "DiaSource %i has NaN value for RA/DEC, "
+                    "dropping from association." %
+                    dia_sources.loc[nan_idx, "diaSourceId"])
+            dia_sources = dia_sources[np.logical_not(nan_mask)]
```
ebellm: You might try a timing test of just this line and see if `dia_sources[~nan_mask]` is faster.

tests/test_association_task.py
```diff
@@ -672,6 +672,26 @@ def test_score_and_match(self):
...
+                for src_idx in range(n_sources)],
+            start_id=0,
+            scatter_arcsec=-1)
+        dia_sources.rename(columns={"coord_ra": "ra",
+                                    "coord_dec": "decl",
+                                    "id": "diaSourceId"},
+                           inplace=True)
+
+        dia_sources.loc[2, "ra"] = np.nan
+        dia_sources.loc[3, "decl"] = np.nan
```
ebellm: this may be silly, but how about a check for both ra and decl as NaN?

python/lsst/ap/association/association.py
morriscb: I doubt there is much difference here so I'll change it to the simpler line.

morriscb:
Updated the ticket. Take a look.

PR #58, opened 2019-10-15
DM-21785: Create HTM index, nDiaSources, and DiaObjectFlags plugin

morriscb:


tests/test_dia_calculation_plugins.py
```diff
@@ -53,39 +56,149 @@ def testCalculate(self):
...
+        """
+        # Test expected pixelIds.
+        diaObject = dict()
+        diaObject["ra"] = 0.
+        diaObject["decl"] = 0.
+        plug = HTMIndexDiaPosition(HTMIndexDiaPositionConfig(),
+                                   "ap_HTMIndex",
+                                   None)
+        plug.calculate(diaObject)
+        self.assertAlmostEqual(diaObject["pixelId"], 131072)
```
mrawls: Can you provide any context about why this (and the subsequent example) is the correct pixelId value?

python/lsst/ap/association/diaCalculationPlugins.py
```diff
@@ -84,6 +88,106 @@ def calculate(self, diaObject, diaSources, **kwargs):
...
             diaObject["ra"] = aveCoord.getRa().asDegrees()
             diaObject["decl"] = aveCoord.getDec().asDegrees()
+            diaObject["radecTai"] = np.max(diaSources["midPointTai"])
+
+
+class HTMIndexDiaPositionConfig(DiaObjectCalculationPluginConfig):
+
+    indexer = IndexerRegistry.makeField(
+        doc='Select the spatial indexer to use within the database.',
+        default='HTM'
```
mrawls: While HTM is a fine default here, you've hardwired "HTM" into the plugin names here. Are there situations where spatial indexers other than HTM might be desired? If so, could you call this something more generic like `SpatialIndexDiaPosition`?

tests/test_dia_calculation_plugins.py
```diff
@@ -53,39 +56,149 @@ def testCalculate(self):
...
         self.assertAlmostEqual(diaObject["decl"], 0.0)
+        self.assertEqual(diaObject["radecTai"], 10)
 
         diaObject = dict()
         diaSources = pd.DataFrame(data={"ra": np.zeros(n_sources),
-                                        "decl": np.linspace(-1, 1, n_sources)})
+                                        "decl": np.linspace(-1, 1, n_sources),
+                                        "midPointTai": np.linspace(0,
+                                                                   n_sources,
+                                                                   n_sources)})
```
mrawls: I'm not sure if these are reasonable timestamps or not, are the units in MJD? Starting at 2450000-whatever might more accurately approximate reality.

python/lsst/ap/association/diaCalculationPlugins.py
morriscb: That would be done by a different plugin.

tests/test_dia_calculation_plugins.py
morriscb: It doesn't matter what the values are. Only that they are round tripped.

tests/test_dia_calculation_plugins.py
morriscb: These are the values produced by the HTM pixelization at the locations requested. The specific values here aren't particularly important, only that they are stable and round trip correctly.

morriscb:
> Please be a little more descriptive in your commit message(s), and consider moving the changes to `MeanDiaPosition` onto a separate commit.
> Are each plugin's output columns documented anywhere? For example, seeing the new `radecTai` output column for `MeanDiaPositionConfig` makes me wonder what the units are and why this name was chosen. It looks like a timestamp, presumably using the TAI standard, which corresponds to each DiaSource, but the name suggests the time itself is somehow a function of the ra and dec, which I don't think is what you mean.

The output columns are described in https://github.com/lsst/dax_ppdb/blob/master/data/ppdb-schema.yaml and https://github.com/lsst/cat.

tests/test_dia_calculation_plugins.py
mrawls: I guess they won't ever be anywhere near too big to store/retrieve properly, so OK.

python/lsst/ap/association/diaCalculationPlugins.py
mrawls: OK, it's a little odd to me then that you have a config option to select the spatial indexer, which implies a different choice could be made. But I don't have a suggestion for a better way to do it. Maybe a check to make sure `indexer == 'HTM'` ?

PR #59, opened 2019-10-21
DM-21167: Migrate ap_association fully to new DiaCalculation plugin system for time-series features.

morriscb:


python/lsst/ap/association/association.py
```diff
@@ -496,95 +302,23 @@ def update_dia_objects(self, dia_objects, updated_obj_ids, exposure, ppdb):
...
-            columns=["used"])
 
         dateTime = exposure.getInfo().getVisitInfo().getDate()
 
         dia_sources = ppdb.getDiaSources(updated_obj_ids, dateTime.toPython(),
                                          return_pandas=True)
-        dia_sources.set_index(["diaObjectId", "diaSourceId"], inplace=True)
+        dia_sources.set_index(
+            ["diaObjectId", "filterName", "diaSourceId"],
+            inplace=True)
```
jdswinbank: Is it necessary to set the index here?

Let me expand on that question a bit, and apologise in advance that this might be due to my Pandas inexperience —

Both `dia_sources` and `dia_objects` are being passed to `diaCalculation.run`, which has requirements on how they are indexed:
```python
        diaObjectCat : `pandas.DataFrame`
            DiaObjects to update values of and append new objects to. DataFrame
            should be indexed on "diaObjectId"
        diaSourceCat : `pandas.DataFrame`
            DiaSources associated with the DiaObjects in diaObjectCat.
            DataFrame should be indexed on
            `["diaObjectId", "filterName", "diaSourceId"]`
```

But all that's happening here as far as I can see is that we load the data, set the index, then pass it to that function.

So that makes me think: rather than making the interface to `diaCalculation.run` brittle (because it depends on the indexes of whatever data we happen to pass to it), why not move the calls to `set_index` to within `diaCalculation.run`? That would make the code here simpler, and it would make `diaCalculation.run` harder to break.

morriscb: Added code to index within diaCalculation and test for already indexed DataFrames.

PR #63, opened 2019-11-18
DM-22234: Remove unnecessary python_future dependency

timj:


timj:
PS @morriscb did you notice that your recent commits on this package have a typo in your author name? Problem with your .gitconfig or deliberate?

PR #65, opened 2019-12-02
DM-22299: Speed up specific diaCalculation plugins using fast pandas functionality

morriscb:


python/lsst/ap/association/diaCalculationPlugins.py
```diff
@@ -1044,8 +1024,5 @@ def calculate(self,
         """
         # Set "delta degrees of freedom (ddf)" to 1 to calculate the unbiased
         # estimator of scatter (i.e. 'N - 1' instead of 'N').
-        def _sigma(df):
-            return np.nanstd(df["totFlux"], ddof=1)
-
         diaObjects.loc[:, "{}TOTFluxSigma".format(filterName)] = \
-            filterDiaSources.apply(_sigma)
+            filterDiaSources.totFlux.agg(np.nanstd)
```
eggls6: check to see if std is nan safe and statistically unbiased / pandas function

morriscb: Checked and the pandas functions are nan-safe and unbiased. I have thus changed to using these functions directly rather than `agg`. Checked for any performance difference using `timeit` and found the run time to be nearly identical with slightly less variance per loop for the pandas builtin.

PR #68, opened 2019-12-18
DM-21886: Create PipelineTask driver for ap_pipe tasks that interact with the APDB

morriscb:


python/lsst/ap/association/diaForcedSource.py
```diff
@@ -129,7 +129,7 @@ def __init__(self, **kwargs):
         self.makeSubtask("forcedMeasurement",
                          refSchema=afwTable.SourceTable.makeMinimalSchema())
 
-    def run(self, dia_objects, expIdBits, exposure, diffim, apdb):
+    def run(self, dia_objects, expIdBits, exposure, diffim):
         """Measure forced sources on the direct and different images,
         calibrate, and store them in the Apdb.
```
isullivan: Does this docstring need to be updated? `run` no longer does anything with the apdb.

python/lsst/ap/association/diaPipe.py
```diff
@@ -0,0 +1,228 @@
...
+        self.makeSubtask("associator")
+        self.makeSubtask("diaForcedSource")
+
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
+        inputs = butlerQC.get(inputRefs)
+        expId, expBits = butlerQC.quantum.dataId.pack("visit_detector",
+                                                      returnMaxBits=True)
+        inputs["ccdExposureIdBits"] = expBits
+
+        outputs = self.run(**inputs)
```
isullivan: Couldn't you just pass `ccdExposureIdBits = expBits` directly, rather than packing it in `inputs`?

python/lsst/ap/association/diaPipe.py
```diff
@@ -0,0 +1,228 @@
...
+        butlerQC.put(outputs, outputRefs)
+
+    @pipeBase.timeMethod
+    def run(self, diaSourceCat, diffIm, exposure, ccdExposureIdBits):
+        """Process DiaSources and DiaObjects.
+
+        Load previous DiaObjects and their DiaSource history. Calibrate the
+        values in the diaSourceCat. Associate new DiaSources with previous
+        DiaObjects. Run forced photometry at the updated DiaObject locations.
+        Store the results in the Apdb.
```
isullivan: Please spell out Apdb here.

python/lsst/ap/association/diaPipe.py
```diff
@@ -0,0 +1,228 @@
...
+
+Currently loads directly from the Apdb rather than pre-loading.
+"""
+
+import os
+
+import lsst.dax.apdb as daxApdb
+import lsst.pex.config as pexConfig
+import lsst.pipe.base as pipeBase
+import lsst.pipe.base.connectionTypes as connTypes
```
isullivan: Could you spell out the full `connectionTypes` (as is used in e.g. pipe_tasks) instead of shortening it to `connTypes`?

python/lsst/ap/association/diaPipe.py
```diff
@@ -0,0 +1,228 @@
...
+from lsst.ap.association import (
+    AssociationTask,
+    DiaForcedSourceTask,
+    LoadDiaCatalogsTask,
+    MapDiaSourceTask,
+    make_dia_object_schema,
+    make_dia_source_schema)
+
+__all__ = ("DiaPipelineConfig",
+           "DiaPipelineTask")
```
isullivan: I think you also need `DiaPipelineConnections` in `__all__`.

python/lsst/ap/association/diaPipe.py
morriscb: Maybe? I saw a few pipe_tasks that used `cT` instead of `connectionTypes` thought this was a good compromise?

python/lsst/ap/association/diaPipe.py
morriscb: Mostly I did this parroting what was happening in calibration.py. They do a similar operation with the same value/variable.

python/lsst/ap/association/diaPipe.py
```diff
@@ -0,0 +1,229 @@
...
+    )
+    diaSourceCat = connTypes.Input(
+        doc="Catalog of DiaSources produced during image differencing.",
+        name="deepDiff_diaSrc",
+        storageClass="SourceCatalog",
+        dimensions=("instrument", "visit", "detector"),
+    )
+    diffIm = connTypes.Input(
+        doc="Difference image on which the DiaSources were detected.",
+        name="deepDiff_differenceExp",
```
kfindeisen: Please use a template instead of hardcoding the "deep". I assume the Gen 3 differencing task, like the old one, will still have multiple types of diffs.
```suggestion
        name="{coaddName}Diff_differenceExp",
```

PR #70, opened 2020-01-06
DM-22727: Add numpy warnings catch to DiaCalculationPlugins

morriscb:


python/lsst/ap/association/diaCalculationPlugins.py
```diff
@@ -559,6 +573,7 @@ def calculate(self,
         **kwargs
             Any additional keyword arguments that may be passed to the plugin.
         """
+
```
isullivan: No blank line after a function docstring.

python/lsst/ap/association/diaCalculationPlugins.py
```diff
@@ -358,12 +362,16 @@ def calculate(self,
             if pTileName not in diaObjects.columns:
                 diaObjects[pTileName] = np.nan
 
-        def _fluxPercentiles(df):
-            pTiles = np.nanpercentile(df["psFlux"], self.config.percentiles)
-            return pd.Series(dict((tileName, pTile)
-                                  for tileName, pTile in zip(pTileNames, pTiles)))
+        with np.warnings.catch_warnings():
+            np.warnings.filterwarnings('ignore', r'All-NaN slice encountered')
```
timj: Can we please use double quotes to be consistent with the rest of the file?  Also, why is this a raw string?

morriscb: Done. The raw string is based on what I saw in examples both online and in our own code: https://github.com/lsst/pipe_tasks/blob/448f008be835c65bd1e4254d7fddb65aacc18f68/python/lsst/pipe/tasks/functors.py#L541

timj: Odd. It can't be doing anything useful can it? Sounds like cargo cult 😄

morriscb:
Hey, @isullivan can you take a look at the new commit and see if that works for you?

jdswinbank:
With apologies for the drive-by comment —

- I don't think there's anything Numpy-specific about this, is there? `numpy.warnings` is just an alias for `warnings`. As such, I'd suggest changing the import, function name, etc, to avoid tying it to Numpy.
- It seems confusing to use `warnings.catch_warnings`, which is a context manager, without a `with` statement. I'm about to get on a plane so I can't check if this works (and maybe I missed something), but it makes me nervous.
- I know you discussed this with Tim, above, but... if you don't need raw strings, please don't use them. Even if other people have elsewhere!

python/lsst/ap/association/diaCalculationPlugins.py
```diff
@@ -58,6 +60,24 @@
            "SigmaDiaTotFlux", "SigmaDiaTotFluxConfig")
 
 
+def catchNumpyWarnings(_func=None, *, warnings=[]):
+    """Decorator for generically catching numpy warnings.
+    """
+    def decoratorCatchWarnings(func):
+        @functools.wraps(func)
+        def wrapperCatchWarnings(*args, **kwargs):
+            np.warnings.catch_warnings()
```
isullivan: Make this `with np.warnings.catch_warnings():` and indent the `for` loop and `return` statement below.

isullivan:
As written, this swallows the specified warnings globally, so you definitely want to still use `with` in your decorator.
As John and Tim mentioned, you don't need the raw strings here.

python/lsst/ap/association/diaCalculationPlugins.py
morriscb: Thanks for catching this. Definitely a bug I introduced while re-writing this.

morriscb:
Verified the changes are working on the ap_verify CI dataset. Will submit another Jenkins just in case. Mind having one last look @isullivan?

isullivan:
My linter still complains about the blank line on 589, which you should remove. Otherwise, it looks good.

PR #71, opened 2020-01-08
DM-22777: Prune diaForcedSources in DiaForcedSourceTask

morriscb:


python/lsst/ap/association/diaForcedSource.py
```diff
@@ -264,3 +266,31 @@ def _calibrate_and_merge(self,
         output_catalog.drop(columns=self.config.dropColumns, inplace=True)
 
         return output_catalog
+
+    def _trim_to_bbox(self, catalog, exposure):
```
kfindeisen: It seems a bit odd to have a `trim_to_bbox` method that takes an exposure. Why not let it take a `Box2I` (more reusable) or call it `trim_to_exposure` (more robust to changes in algorithm, pixel vs. celestial coordinates, etc)?

python/lsst/ap/association/diaForcedSource.py
```diff
@@ -264,3 +266,31 @@ def _calibrate_and_merge(self,
...
+        minX = bbox.getMinX()
+        minY = bbox.getMinY()
+        maxX = bbox.getMaxX()
+        maxY = bbox.getMaxY()
+
+        xS = catalog.loc[:, "x"]
+        yS = catalog.loc[:, "y"]
+
+        return catalog[np.logical_and(np.logical_and(minX < xS, maxX > xS),
+                                      np.logical_and(minY < yS, maxY > yS))]
```
kfindeisen: `Box2D` has a vector `contains` method, so I think you can simplify this to something like:

```
return catalog[bbox.contains(xS, yS)]
```

The loop over `xS` and `yS` is [done in C++](https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html#vectorizing-functions), so this should also be faster than the current code.

tests/test_diaForcedSource.py
```diff
@@ -59,6 +59,13 @@ def create_test_dia_objects(n_points, wcs, startPos=100):
         src['id'] = src_idx
         src.setCoord(wcs.pixelToSky(startPos + src_idx,
                                     startPos + src_idx))
+
+    # Add one point outside of the CCD.
+    src = objects.addNew()
+    src['id'] = 10000000
+    src.setCoord(wcs.pixelToSky(-100000,
+                                -100000))
```
kfindeisen: Actually, one more suggestion: consider testing points that are out-of-bounds in only X or only Y, or some that are *just* out-of-bounds (especially points that would be treated differently by `Box2I` vs. `Box2D`).

python/lsst/ap/association/diaForcedSource.py
morriscb: Changed to `trim_to_exposure`

python/lsst/ap/association/diaForcedSource.py
morriscb: Thanks, didn't see the vectorized bbox.contains.

PR #81, opened 2020-05-15
Close open file descriptors (psutil check fixed from DM-24907)

brianv0:


tests/test_loadDiaCatalogs.py
```diff
@@ -264,7 +264,7 @@ def setUp(self):
                                   self.dateTime)
 
     def tearDown(self):
-        del self.tmp_file
+        os.close(self.db_file_fd)
         os.remove(self.db_file)
         del self.db_file
```
timj, line 269: I don't think this del does anything useful.

brianv0: I'll remove it

PR #82, opened 2020-05-21
DM-25010: Make Alert serialization optional in diaPipe.

morriscb:


python/lsst/ap/association/diaPipe.py
```diff
@@ -122,6 +122,11 @@ class DiaPipelineConfig(pipeBase.PipelineTaskConfig,
         target=PackageAlertsTask,
         doc="Subtask for packaging Ap data into alerts.",
     )
+    doSerializeAlerts = pexConfig.Field(
```
jdswinbank, line 125: Is `doSerializeAlerts` the write term here?

I mean, it's correct, but it seems to add an unnecessary third term beyond “package” (on the basis that it's `PackageAlertsTask`, rather than `SerializeAlertsTask` that we're referring to) and “write” (on the basis that it's really writing out the alerts that the user cares about)

It's not a big deal, it just feels like we're using three terms which all kind-of-but-not-quite mean the same thing, where one (or _maybe_ two) would do fine.

tests/test_diaPipe.py
```diff
@@ -36,7 +36,7 @@
 class TestDiaPipelineTask(unittest.TestCase):
 
     @classmethod
-    def _makeDefaultConfig(cls):
+    def _makeDefaultConfig(cls, doAlerts=False):
```
jdswinbank, line 39: Would you consider using `doSerializeAlerts` here? (Or whatever you change that to, if you think my comment on naming is helpful). Two reasons: one for consistency, and the other because `doAlerts` is not very descriptive (“do alerts what?”).

python/lsst/ap/association/diaPipe.py
morriscb: Changed everything to PackageAlerts.

tests/test_diaPipe.py
```diff
@@ -117,7 +129,10 @@ def testRun(self):
             subtasks.dpddifier.run.assert_called_once()
             subtasks.associator.run.assert_called_once()
             subtasks.diaForcedSource.run.assert_called_once()
-            subtasks.alertPackager.run.assert_called_once()
+            if doAlerts:
+                subtasks.alertPackager.run.assert_called_once()
+            else:
+                subtasks.alertPackager.run.assert_not_called()
```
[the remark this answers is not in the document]
morriscb: Picked the first option

jdswinbank:
Hey Chris — thanks for these changes.

Looking at the changes you made to `TestDiaPipelineTask`, I realised that I hadn't previously understood what it was doing. Unfortunately, that means that the change I suggested (specifically,  the `self.assertFalse(hasattr(subtasks, 'alertPackager'))`) is not very useful — it's only demonstrating that the test code works, not that the task itself is doing the right thing.

I spent a few minutes playing about with the code and came up with these changes: https://github.com/lsst/ap_association/commit/db2c0c281b3777a77583b4e615d92d418ecc1f8b. I think they make the test a bit simpler and more useful than it is now. What do you think?

morriscb:
Sure, that looks good to me. Since it's on your branch, do you mind appending it to the end of this ticket branch? You're much better with git after all :D

jdswinbank:
Appended!

PR #83, opened 2020-05-22
DM-25040: ap_association uses physical filter in Gen 3

kfindeisen:
This PR fixes a bug that caused `MapDiaSourceTask` to sometimes report observatory-specific filters, which would then get propagated to table columns. It updates all `ap_association` code to standardize filter names.

### DIFF
--- a/python/lsst/ap/association/afwUtils.py
+++ b/python/lsst/ap/association/afwUtils.py
@@ -805,7 +805,7 @@ def get_ccd_visit_info_from_exposure(exposure):
     #  flux zero point error ``counts``]
     values = {'ccdVisitId': visit_info.getExposureId(),
               'ccdNum': exposure.getDetector().getId(),
-              'filterName': filter_obj.getName(),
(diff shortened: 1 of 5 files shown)


python/lsst/ap/association/afwUtils.py
```diff
@@ -805,7 +805,7 @@ def get_ccd_visit_info_from_exposure(exposure):
     #  flux zero point error ``counts``]
     values = {'ccdVisitId': visit_info.getExposureId(),
               'ccdNum': exposure.getDetector().getId(),
-              'filterName': filter_obj.getName(),
+              'filterName': filter_obj.getCanonicalName(),  # use abstract filter
```
kfindeisen, line 808: `get_ccd_visit_info_from_exposure` doesn't seem to be used anywhere; can it be removed?

morriscb: Yes it can. It's left over from a previous iteration of the Apdb and, as you point out, is not used anywhere.

PR #85, opened 2020-06-05
DM-25280: Adapt to new alert_packet API

jdswinbank:


python/lsst/ap/association/packageAlerts.py
```diff
@@ -42,7 +42,9 @@ class PackageAlertsConfig(pexConfig.Config):
         dtype=str,
         doc="Schema definition file for the avro alerts.",
         default=os.path.join(getPackageDir("alert_packet"),
```
timj, line 44: I think whilst this solution fixes the immediate problem it doesn't seem to me like the right solution long term. If alert_packet is installed properly then you should be using pkg_resources to locate the schemas and if they are in a zip file you will need to get a handle to the contents and not a file itself. I guess this works because of the other fix in alert_packet where you don't us python setup.py at all but copy everything over without involving setuptools.

jdswinbank: I agree with all of those statements.

PR #89, opened 2020-06-19
DM-24325: Include CCDData cutouts in alert packets

morriscb:


python/lsst/ap/association/packageAlerts.py
```diff
@@ -165,6 +177,80 @@ def createDiaSourceBBox(self, bboxSize):
...
+        photoCalib : `lsst.afw.image.PhotoCalib`
+            Calibrate object of the image the cutout is cut from.
+
+        Returns
+        -------
+        ccdData : `astropy.nddata.CCDData`
+            CCDData object storing the calibrate information from the input
+            difference or template image.
+        """
+        cutOutMinX = cutout.getBBox().minX - 1
```
isullivan, line 198: Please add a comment explaining the additional `-1` here.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -165,6 +177,80 @@ def createDiaSourceBBox(self, bboxSize):
...
+        cutOutMinY = cutout.getBBox().minY - 1
+        center = cutout.getWcs().skyToPixel(sphPoint)
+        calibCutout = photoCalib.calibrateImage(cutout.getMaskedImage())
+
+        cutoutWcs = wcs.WCS(naxis=2)
+        cutoutWcs.array_shape = (cutout.getBBox().getWidth(),
+                                 cutout.getBBox().getWidth())
+        cutoutWcs.wcs.crpix = [center.x - cutOutMinX, center.y - cutOutMinY]
```
isullivan, line 206: Why is the crpix at the corner rather than the center of the cutout?

python/lsst/ap/association/packageAlerts.py
```diff
@@ -165,6 +177,80 @@ def createDiaSourceBBox(self, bboxSize):
...
+        cutoutWcs.wcs.crval = [sphPoint.getRa().asDegrees(),
+                               sphPoint.getDec().asDegrees()]
```
isullivan, line 208: Isn't this crval actually for the center of the bbox?

python/lsst/ap/association/packageAlerts.py
```diff
@@ -165,6 +177,80 @@ def createDiaSourceBBox(self, bboxSize):
             bbox = geom.Extent2I(bboxSize, bboxSize)
         return bbox
 
+    def createCcdDataCutout(self, cutout, sphPoint, photoCalib):
```
isullivan, line 180: I feel `sphPoint` describes the data type more than what the parameter is. Maybe something like `centerPoint` instead?

python/lsst/ap/association/packageAlerts.py
```diff
@@ -165,6 +177,80 @@ def createDiaSourceBBox(self, bboxSize):
...
+            Point on sky to approximate the Wcs.
+
+        Returns
+        -------
+        localMatrix : `numpy.ndarray`
+            Matrix representation the local wcs approximation with units
+            degrees.
+        """
+        localGnomonicWcs = afwGeom.makeSkyWcs(
+            center, skyCenter, [[self._scale, 0], [0, self._scale]])
```
isullivan, line 247: I think this would be more readable if you first defined a variable such as `cdMatrix = [[self._scale, 0], [0, self._scale]]` and then used that here.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -165,6 +177,80 @@ def createDiaSourceBBox(self, bboxSize):
...
+            unit=u.nJy)
+
+    def makeLocalTransformMatrix(self, wcs, center, skyCenter):
+        """Create a local, linear approximation of the wcs transformation
+        matrix.
+
+        The approximation is created as if the center is at RA=0, DEC=0. All
+        comparing x,y coordinate are relative to the position of center. Matrix
+        is initially calculated with units arcseconds and then converted to
+        radians. This yields higher precision results due to quirks in AST.
```
isullivan, line 229: "converted to radians". Do you mean "converted to degrees"?

tests/test_packageAlerts.py
```diff
@@ -288,22 +299,56 @@ def testCreateBBox(self):
...
+        packageAlerts = PackageAlertsTask()
+
+        ccdData = packageAlerts.createCcdDataCutout(
+            self.exposure,
+            self.exposure.getWcs().getSkyOrigin(),
+            self.exposure.getPhotoCalib())
+        calibExposure = self.exposure.getPhotoCalib().calibrateImage(
+            self.exposure.getMaskedImage())
+
+        self.assertTrue(np.allclose(ccdData.wcs.wcs.cd,
```
isullivan, line 315: I think `self.assertFloatsAlmostEqual(ccdData.wcs.wcs.cd, self.cutoutWcs.wcs.cd)` might give more informative error messages if the test fails. The downside is you would have to have `TestPackageAlerts` inherit from `lsst.utils.tests.TestCase` instead of `unittest.TestCase`.

If you end up switching the assert here, also change the other `self.assertTrue(np.allclose(...` throughout this file.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -106,6 +113,7 @@ def run(self,
         self._patchDiaSources(diaSourceCat)
```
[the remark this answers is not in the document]
morriscb: Eventually will so I'm going to leave it for now.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -117,7 +125,11 @@ def run(self,
...
             cutoutBBox = self.createDiaSourceBBox(diaSource["bboxSize"])
-            diffImCutout = diffIm.getCutout(sphPoint, cutoutBBox)
+            diffImCutout = self.createCcdDataCutout(
+                diffIm.getCutout(sphPoint, cutoutBBox),
+                sphPoint,
+                diffImPhotoCalib)
+
             templateCutout = None
             # TODO: Create alertIds DM-24858
             alertId = diaSource["diaSourceId"]
```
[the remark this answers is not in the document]
morriscb: Removed. Should have been removed in a previous ticket but was not.

python/lsst/ap/association/packageAlerts.py
morriscb: So, CCDData doesn't store absolute pixel locations. So when you create a dataset with it the "bottom left" pixel is always [1, 1]. Thus we center the crpix such that it's correctly within this newly shifted cutout and at the location of the DiaSource in the image. There is a possible future feature in astropy to add an origin to the CCDData object but it is currently not a priority.

python/lsst/ap/association/packageAlerts.py
morriscb: It's the center of the DiaSource, see above.

python/lsst/ap/association/packageAlerts.py
morriscb: Changed to skyCenter

python/lsst/ap/association/packageAlerts.py
timj: I think "not a priority" is right for them but if we want to add that feature they seem like they would be more than happy to accept a patch. We have been in discussion for years to add pixel origin to NDData.

PR #90, opened 2020-06-24
DM-25164: Add diaForceSources to diaPipe loader and to alert packets

morriscb:


python/lsst/ap/association/diaPipe.py
```diff
@@ -239,9 +241,14 @@ def run(self, diaSourceCat, diffIm, exposure, ccdExposureIdBits):
             exposure.getInfo().getVisitInfo().getDate().toPython())
         self.apdb.storeDiaForcedSources(diaForcedSources)
         if self.config.doPackageAlerts:
+            diaForcedSources.set_index("diaObjectId", drop=False, inplace=True)
+            if len(loaderResult.diaForcedSources) > 1:
+                diaForcedSources = pd.concat([diaForcedSources,
+                                              loaderResult.diaForcedSources])
```
spenczar, line 247: I'm kind of surprised by the order here, I'd have guessed that you need to `set_index` _after_ doing the `concat` with `loaderResult`. Maybe this is fine, I don't really know, just mentioning it.

spenczar: Oh yeah, I see the same index is set in `loadDiaForcedSources`, which I guess is where this value comes from, right?

python/lsst/ap/association/loadDiaCatalogs.py
```diff
@@ -192,6 +197,35 @@ def loadDiaSources(self, diaObjects, pixelRanges, dateTime, apdb):
...
         return diaSources.replace(to_replace=[None], value=np.nan)
 
+    @pipeBase.timeMethod
+    def loadDiaForcedSources(self, diaObjects, dateTime, apdb):
+        """Load DiaObjects from the Apdb based on their HTM location.
+
+        Parameters
+        ----------
+        pixelRanges : `tuple` [`int`]
+            Ranges of pixel values that cover region of interest.
```
spenczar, line 207: This looks out of sync with the function params

tests/test_packageAlerts.py
```diff
@@ -167,7 +167,45 @@ def makeDiaSources(nSources, diaObjectIds, exposure):
...
+    Parameters
+    ----------
+    nSources : `int`
+        Number of sources to create.
+    diaObjectIds : `numpy.ndarray`
+        Integer Ids of diaobjects to "associate" with the DiaSources.
+    exposure : `lsst.afw.image.Exposure`
+        Exposure to create sources over.
+    pixelator : `lsst.sphgeom.HtmPixelization`
+        Object to compute spatial indicies from.
```
spenczar, line 182: This isn't in the function signature, maybe out of date

python/lsst/ap/association/loadDiaCatalogs.py
```diff
@@ -192,6 +197,35 @@ def loadDiaSources(self, diaObjects, pixelRanges, dateTime, apdb):
...
+            DiaObjects loaded from the Apdb that are within the area defined
+            by ``pixelRanges``.
+        """
+        if len(diaObjects) == 0:
+            # If no diaObjects are available return an empty DataFrame with
+            # the the column used for indexing later in AssociationTask.
+            diaForcedSources = pd.DataFrame(columns=["diaObjectId"])
+        else:
+            diaForcedSources = apdb.getDiaForcedSources(
+                diaObjects.loc[:, "diaObjectId"],
```
spenczar, line 223: Should the object IDs be trimmed down to just unique values? Maybe it's already guaranteed to be unique, but what happens if there are object IDs that appear multiple times, are forced sources duplicated in the result here?

python/lsst/ap/association/diaPipe.py
morriscb: Yeah, what you're seeing here was an attempt to get around the fact that when joined, dataframes can do weird things with the types. Now that you mention it, I'm not sure the data ends up sorted here. I'll switch back to an appended with a specific "sort" to make this more explicit.

python/lsst/ap/association/loadDiaCatalogs.py
morriscb: So, the daiObjectIds are duplicated for sure but the row values are unique so I don't think there will be any duplicated data here. I believe pandas groups together the duplicated values, but what I can do is specifically add the diaForcedSourceId here to make sure this is what is happening and get an ordering for the diaForcedSources in the alert packet.

python/lsst/ap/association/diaPipe.py
morriscb: I moved the diaForcedSources indexing into the diaForcedSource task to keep it more compartmentalized.

python/lsst/ap/association/loadDiaCatalogs.py
morriscb: Sorry! Copy paste bites me again.

morriscb:
Hey @spenczar, could you take a quick look, please, and see if I have responded to your review adequately. Thanks!

spenczar:
@morriscb I think it all makes sense, but I don't see any commits pushed, just to be clear.

morriscb:
Oops, sorry. Just pushed now.

python/lsst/ap/association/loadDiaCatalogs.py
```diff
@@ -192,6 +197,41 @@ def loadDiaSources(self, diaObjects, pixelRanges, dateTime, apdb):
...
+        dataTime : `datetime.datetime`
+            Time of the current visit
+        apdb : `lsst.dax.apdb.Apdb`
+            Database connection object to load from.
+
+        Returns
+        -------
+        diaObjects : `pandas.DataFrame`
+            DiaObjects loaded from the Apdb that are within the area defined
+            by ``pixelRanges``.
```
spenczar, line 218: `pixelRanges` is out of date

python/lsst/ap/association/loadDiaCatalogs.py
```diff
@@ -192,6 +197,41 @@ def loadDiaSources(self, diaObjects, pixelRanges, dateTime, apdb):
...
+    @pipeBase.timeMethod
+    def loadDiaForcedSources(self, diaObjects, dateTime, apdb):
+        """Load DiaObjects from the Apdb based on their HTM location.
+
+        Parameters
+        ----------
+        diaObjects : `pandas.DataFrame`
+            DiaObjects loaded from the Apdb that are within the area defined
+            by ``pixelRanges``.
+        dataTime : `datetime.datetime`
```
spenczar, line 209: typo: `dataTime` -> `dateTime`

PR #96, opened 2020-07-31
Use iat accessor for pandas 1.1.0 compatibility

brianv0:
Pandas 1.1.0 now throws a KeyError in parts of the code. The fix is to use the `iat` accessor.


brianv0:
This builds passes jenkins for 1.1.0 as well as 1.0.3 environments.

PR #99, opened 2020-09-23
DM-26783: DiaForcedSource on associated DiaObject off frame

morriscb:


tests/test_diaForcedSource.py
```diff
@@ -37,7 +37,9 @@
 
 
 def create_test_dia_objects(n_points, wcs, startPos=100):
-    """Create dummy DIASources or DIAObjects for use in our tests.
+    """Create dummy DIASources or DIAObjects for use in our tests. Adds
+    Three test sources outsize of the ccd area.
```
isullivan, line 41: "outside" of the ccd area?

tests/test_diaForcedSource.py
```diff
@@ -37,7 +37,9 @@
...
+
```
isullivan, line 42: Not in the currently changed code, but I notice below that two of the outside sources are given `src['id'] = 10000001`. Is that a bug?

tests/test_diaForcedSource.py
```diff
@@ -202,7 +211,7 @@ def testRun(self):
 
         # Should be number of test objects minus one as one object is purposely
         # outside of the ccd area.
-        self.assertEqual(len(dia_forced_sources), len(self.testDiaObjects) - 3)
+        self.assertEqual(len(dia_forced_sources), len(self.testDiaObjects) - 2)
```
isullivan, line 214: I'm confused about this line and the comment. It looks like you're adding 3 sources outside the ccd in `create_test_dia_objects`, yet the comment says one and the comparison here looks like two.
After playing with the code a little, it looks like the selection of five sources on line 193 is including two of the three outside sources, and omitting the third. If I expand that to `[1:7]` toinclude 6 sources and also fix the source id of the third outside source, then this line needs to be `len(self.testDiaObjects) - 1` to pass, which matches the comment. I'm pretty sure I'm missing something here, though. Could you please double-check the logic, and either expand the comment to explain why there is a `-2` here, or modify the code if it is incorrect in some way?

isullivan (changes requested):
Please double-check the logic in `TestDiaForcedSource.testRun` and provide a more detailed description of where the numbers in the length test come from.

tests/test_diaForcedSource.py
morriscb: Thanks for the find, fixed.

tests/test_diaForcedSource.py
morriscb: Moved them to the setUp section and added a bit more clarity in the comments.

tests/test_diaForcedSource.py
morriscb: Made a variable in setup and added more comments for clarity (hopefully).

tests/test_diaForcedSource.py
```diff
@@ -175,17 +159,44 @@ def setUp(self):
...
+        # x outside
+        src = self.testDiaObjects.addNew()
+        src['id'] = 10000002
+        src.setCoord(self.wcs.pixelToSky(-100000,
+                                         100))
+        # Ids of objects that were "updated" during "ap_association"
+        # processing.
+        self.updatedTestIds = np.array([1, 2, 3, 4, 10000001], dtype=np.uint64)
+        # Expecdted number of sources is the number of updated ids plus
+        # plus any that are within the CCD footprint but are not in the
```
isullivan, line 183: Nitpick: "plus plus"

isullivan (approved):
This looks much more clear now, thank you!

PR #100, opened 2020-10-16
DM-25990: Reprocess HSC COSMOS medium dataset with ap_pipe

mrawls:
As in DM-25746, this is a temporary workaround to enable R2 and I2 processing with ap_pipe. All the APDB filter name columns presently require a single character, and the DiaForcedSource table is no exception.

### DIFF
--- a/python/lsst/ap/association/diaForcedSource.py
+++ b/python/lsst/ap/association/diaForcedSource.py
@@ -25,8 +25,6 @@ locations.
 
 __all__ = ["DiaForcedSourceTask", "DiaForcedSourcedConfig"]
 
-import numpy as np
-
 import lsst.afw.table as afwTable
 from lsst.daf.base import DateTime
 import lsst.geom as geom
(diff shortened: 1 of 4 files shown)


python/lsst/ap/association/diaForcedSource.py
```diff
@@ -277,7 +277,8 @@ def _calibrate_and_merge(self,
         midPointTaiMJD = visit_info.getDate().get(system=DateTime.MJD)
         output_catalog["ccdVisitId"] = ccdVisitId
         output_catalog["midPointTai"] = midPointTaiMJD
-        output_catalog["filterName"] = diff_exp.getFilter().getCanonicalName()
+        # TODO DM-21333: Remove [0] (first character only) workaround
+        output_catalog["filterName"] = diff_exp.getFilter().getCanonicalName()[0]
```
isullivan, line 281: DM-21333 is reported as 'Done', so you may want to replace the TODO with when RFC-730 is "Implemented"

isullivan: Could you add a comment about what quantity you are trying to retrieve here (i.e. band?, physical filter?), since that will help the conversion effort to the new filter.

isullivan (approved):

python/lsst/ap/association/diaForcedSource.py
mrawls: Good point, I will reference DM-27170 instead, which appears to be one of the final pieces of RFC-730. This issue appears in a few other places in ap_association as well, so I will update those accordingly. The goal here is to retrieve a single-character representation of the band.

PR #101, opened 2020-11-17
DM-26592: Create “afterburner” to analyse the results of AP processing with fakes included

morriscb:


python/lsst/ap/association/diaPipe.py
```diff
@@ -95,6 +95,19 @@ class DiaPipelineConnections(pipeBase.PipelineTaskConnections,
         storageClass="Config",
         dimensions=("instrument", "visit", "detector"),
     )
+    associatedDiaSources = connTypes.Output(
+        doc="Optional output storing the DiaSource catalog after matching and "
+            "SDMification.",
```
kgabor, line 100: Please avoid the `SDMification` jargon. I could not find any usable definition anywhere in the docs, appears occasionally in jira tickets only. Point to a relevant DMTN if there is any or the typical task (MapDiaSourceTask ?) that produces this data model.

kgabor (approved):

PR #102, opened 2021-01-02
DM-27170: Deprecate and phase out afw::image::Filter

parejkoj:


data/apdb-ap-pipe-schema-extra.yaml
```diff
@@ -612,10 +612,6 @@ columns:
   type: CHAR
   nullable: false
   description: Name of stored filter.
-- name: filterId
-  type: INT
-  nullable: false
-  description: id number of stored filter.
```
kfindeisen, line 618: Even if it wasn't being used, I think any kind of schema change in the APDB could break things, so is @morriscb ok with this?

python/lsst/ap/association/diaForcedSource.py
```diff
@@ -277,9 +277,7 @@ def _calibrate_and_merge(self,
         midPointTaiMJD = visit_info.getDate().get(system=DateTime.MJD)
         output_catalog["ccdVisitId"] = ccdVisitId
         output_catalog["midPointTai"] = midPointTaiMJD
-        # TODO DM-27170: fix this [0] workaround which gets a single character
-        # representation of the band.
-        output_catalog["filterName"] = diff_exp.getFilter().getCanonicalName()[0]
+        output_catalog["filterName"] = diff_exp.getFilterLabel().bandLabel
```
kfindeisen, line 280: Can you include a fallback for when `diff_exp.getFilterLabel() == None` (or equivalently `diff_exp.hasFilterLabel() == False`)? `""` seems like a reasonable value for that case (though that would also require special-casing physical-only filters). Same with the other assignments of `bandLabel`.

tests/test_diaForcedSource.py
```diff
@@ -67,15 +66,6 @@ def create_test_dia_objects(n_points, wcs, startPos=100):
 class TestDiaForcedSource(unittest.TestCase):
 
     def setUp(self):
-        # CFHT Filters from the camera mapper.
-        self.filter_names = ["u", "g", "r", "i", "z"]
```
kfindeisen, line 71: Don't you still need `self.filter_names`?

kfindeisen (commented):
I'm a bit worried about the removal of `filterId` (which should have been its own commit, IMO), so I'd like that cleared up before I approve.

data/apdb-ap-pipe-schema-extra.yaml
parejkoj: On slack, @morriscb said "The later isn’t used anymore I don’t think so that’s fine": https://lsstc.slack.com/archives/C2B6X08LS/p1608158086072200

python/lsst/ap/association/diaForcedSource.py
parejkoj: Why shouldn't this just fail if there is no bandLabel? That seems like a recipe for subtly-buggy future catalogs.

tests/test_diaForcedSource.py
parejkoj: It appears it was never used in this particular test. The only test that still needs it is `test_association_task.py`

kfindeisen (approved):

PR #103, opened 2021-02-04
DM-28092: Document that DiaPipeTask can only handle specific bands

morriscb:


python/lsst/ap/association/diaPipe.py
```diff
@@ -125,6 +150,13 @@ class DiaPipelineConfig(pipeBase.PipelineTaskConfig,
         doc="Database connection for storing associated DiaSources and "
             "DiaObjects. Must already be initialized.",
     )
+    validBands = pexConfig.ListField(
+        dtype=str,
+        default=["u", "g", "r", "i", "z", "y"],
+        doc="List of bands that are  valid for AP processing. To process a "
+            "band not on this list, the appropriate, band specific columns "
```
parejkoj, line 157: No comma after "appropriate", I think.

python/lsst/ap/association/diaPipe.py
```diff
@@ -125,6 +150,13 @@ class DiaPipelineConfig(pipeBase.PipelineTaskConfig,
...
+            "must be added to the Apdb schema in dax_apdb.",
```
parejkoj, line 158: Apdb all caps nor no caps?

python/lsst/ap/association/diaPipe.py
```diff
@@ -109,6 +109,31 @@ def __init__(self, *, config=None):
...
+        datasetRefMap : `NamedKeyDict`
+            Mapping from dataset type to a `set` of
+            `lsst.daf.butler.DatasetRef` objects
+        """
+        refs = datasetRefMap[self.diffIm.name]
+        for ref in refs:
+            if ref.dataId["band"] not in self.config.validBands:
+                raise ValueError(
+                    f"Requested '{ref.dataId['band']}' not in validBands list "
+                    "in DiaPipelineTask config. To process bands not in the "
```
parejkoj, line 131: I would just say "not in DiaPipelineConfig.validBands"

python/lsst/ap/association/diaPipe.py
```diff
@@ -109,6 +109,31 @@ def __init__(self, *, config=None):
...
+                    "standard Rubin set, ugrizy, you must add the band to the "
```
parejkoj, line 132: Parentheses instead of commas around "ugrizy"?

parejkoj (commented):
Is there a way to write a test for this?

I think you want to squash the commits?

python/lsst/ap/association/diaPipe.py
```diff
@@ -109,6 +109,31 @@ def __init__(self, *, config=None):
         if not config.doWriteAssociatedSources:
             self.outputs.remove("associatedDiaSources")
 
+    def adjustQuantum(self, datasetRefMap: pipeBase.InputQuantizedConnection):
+        """Override to make adjustments to `lsst.daf.butler.DatasetRef` objects
+        in the `lsst.daf.butler.core.Quantum` during the graph generation stage
+        of the activator.
+
+        This implementation checks to make sure that the filters in the dataset
+        are compatible with AP processing as set by the Apdb/DPDD schema.
```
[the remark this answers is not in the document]
morriscb: Because I'm using the name of the task in dax_apdb. Not sure it there is a standard for how to name stuff like so I will go with what the interface object is called.

python/lsst/ap/association/diaPipe.py
morriscb: See above

morriscb:
Fixed the commit history.

I've got a ticket, DM-22743, to implement runQuantum unittests. I can do the unittest for that.

python/lsst/ap/association/diaPipe.py
```diff
@@ -125,6 +162,13 @@ class DiaPipelineConfig(pipeBase.PipelineTaskConfig,
         doc="Database connection for storing associated DiaSources and "
             "DiaObjects. Must already be initialized.",
     )
+    validBands = pexConfig.ListField(
+        dtype=str,
+        default=["u", "g", "r", "i", "z", "y"],
+        doc="List of bands that are  valid for AP processing. To process a "
```
parejkoj, line 168: double space between `are  valid`

python/lsst/ap/association/diaPipe.py
```diff
@@ -109,6 +109,43 @@ def __init__(self, *, config=None):
...
+        """
+        refs = datasetRefMap[self.diffIm.name]
+        for ref in refs:
+            if ref.dataId["band"] not in self.config.validBands:
+                raise ValueError(
+                    f"Requested '{ref.dataId['band']}' not in "
+                    "DiaPipelineConfig.validBands. To process bands not in the "
+                    "standard Rubin set (ugrizy) you must add the band to the "
+                    "validBands list in DiaPipelineConfig and add the "
+                    "appropriate columns to the Apdb schema.")
```
parejkoj, line 146: I think this ValueError should include the dataId and/or ref that failed. Otherwise the user might not know exactly where that particular band came from.

parejkoj (approved):
One more typo and an addition to the ValueError.

Please squash the commits and cleanup the commit message (a lot of unnecessary "debug X" in there).

Please do add a test for this when you implement runQuantum tests. Maybe add a note to that ticket as a reminder?

python/lsst/ap/association/diaPipe.py
morriscb: I don't think that would be too helpful. As is it would just report the first dataId it encountered which would possibly just prompt the user to remove them one after the other vs removing them all.

morriscb: I'll add a --show-data to the error message.

morriscb: Actually, re-reading my error message. I'm fine with keeping it as is.

PR #105, opened 2021-03-03
DM-28940: Create functor for pix^2->arcsec^2 unit conversion

morriscb:


morriscb:
Going to save adding the functor to the yaml for a later implementation ticket DM-29147.

timothydmorton (commented):
Looks good!

PR #106, opened 2021-03-12
DM-28555: Add verbosity to ApPipe and DiaPipe DB errors

morriscb:


python/lsst/ap/association/association.py
```diff
@@ -129,13 +129,36 @@ def run(self,
 
         diaObjects = diaObjects.append(matchResult.new_dia_objects,
                                        sort=True)
+        # if len(diaObjects) > 0:
+        #     dups = diaObjects.iloc[[0, -1]]
+        #     diaObjects = diaObjects.append(dups, sort=True)
```
isullivan, line 134: Delete this commented-out code.

python/lsst/ap/association/loadDiaCatalogs.py
```diff
@@ -97,11 +98,19 @@ def run(self, exposure, apdb):
               ``diaObjectId``, ``filterName``, ``diaSourceId`` columns.
               (`pandas.DataFrame`)
         """
+        visit_info = exposure.getInfo().getVisitInfo()
```
isullivan, line 101: Shouldn't this be `visitInfo`?

python/lsst/ap/association/association.py
```diff
@@ -129,13 +129,36 @@ def run(self,
...
+        if diaObjects.index.has_duplicates:
+            self.log.warn(
+                "Duplicate DiaObjects created after association. This may "
+                "cause downstream pipeline issues. Dropping duplicated rows.")
+            # Drop duplicates via index and keep the first appearance.
+            diaObjects = diaObjects.groupby(diaObjects.index).first()
```
isullivan, line 140: Here, and throughout these changes where you use `groupby`: does this change the ordering of the diaObjects?

isullivan (approved):
Looks good. My main question is whether the grouping you do to find and drop duplicates changes the order of the diaObjects, and if it does, whether that matters.

python/lsst/ap/association/association.py
morriscb: Thanks for the catch. Was keeping this in until the decision was made on warn vs raise for finding duplicates. Will take it out.

python/lsst/ap/association/association.py
morriscb: It might, but since the dataFrames are indexed and continue to be so after this interaction it won't matter. I rarely use iloc and only do so after matching it to the proper location within the arrays.

isullivan (approved):
Looks good!

python/lsst/ap/association/association.py
```diff
@@ -127,15 +127,34 @@ def run(self,
...
+        # Test for DiaSource duplication first. If duplicates are found,
+        # this likely means this is duplicate data being processed and sent
+        # to the Apdb.
         mergedDiaSourceHistory = diaSourceHistory.append(diaSources, sort=True)
+        if mergedDiaSourceHistory.index.has_duplicates:
+            raise RuntimeError(
+                "Duplicate DiaSources found after association and merging "
+                "with history. This is likely due to re-running data with an "
+                "already populated Apdb. If this was not the case then there "
+                "was a failure in Association which should not happen. "
```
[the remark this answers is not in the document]
morriscb: I expanded a bit but left the heart of what you suggested.

PR #108, opened 2021-04-05
DM-28961: DiaSource centroids outside of image causing crash in cuttout creation in PackageAlerts.

morriscb:


python/lsst/ap/association/packageAlerts.py
```diff
@@ -186,24 +190,49 @@ def createDiaSourceBBox(self, bboxSize):
...
 
         Parameters
         ----------
-        cutout : `lsst.afw.image.ExposureF`
-            Cutout to convert.
+        image : `lsst.afw.image.ExposureF`
+            Image to pull cutout from.
         skyCenter : `lsst.geom.SpherePoint`
             Center point of DiaSource on the sky.
+        bbox : `lsst.geom.Extent2I`
```
isullivan, line 202: Isn't it a `BBox` and not and `Extent2I`?

tests/test_packageAlerts.py
```diff
@@ -357,7 +357,9 @@ def testCreateCcdDataCutout(self):
         ccdData = packageAlerts.createCcdDataCutout(
             self.exposure,
             self.exposure.getWcs().getSkyOrigin(),
-            self.exposure.getPhotoCalib())
+            self.exposure.getBBox().getDimensions(),
+            self.exposure.getPhotoCalib(),
+            1234)
```
isullivan, line 362: Please define `1234` as a variable (something like `diaSourceId = 1234) and pass that in, instead of a number. Also for the rest of this file where you are passing in a source ID.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -186,24 +190,49 @@ def createDiaSourceBBox(self, bboxSize):
...
+        # Catch errors in retrieving the cutout.
+        try:
+            cutout = image.getCutout(skyCenter, bbox)
+        except InvalidParameterError:
+            point = image.getWcs().skyToPixel(skyCenter)
+            imBBox = image.getBBox()
+            self.log.warn(
+                "Failed to retrieve cutout from image for DiaSource with "
+                "id=%i. InvalidParameterError thrown during cutout creation. "
+                "Returning `None` for cuttout."
```
isullivan, line 225: `cuttout` -> `cutout`

python/lsst/ap/association/packageAlerts.py
```diff
@@ -186,24 +190,49 @@ def createDiaSourceBBox(self, bboxSize):
...
+                % srcId)
+            if not geom.Box2D(image.getBBox()).contains(point):
+                self.log.warn(
+                    "DiaSource centroid lies at pixel (%.2f, %.2f) "
```
isullivan, line 229: Please include the source ID in this warning as well.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -186,24 +190,49 @@ def createDiaSourceBBox(self, bboxSize):
...
+                    "which is outside the Exposure with bounding box "
+                    "((%i, %i), (%i, %i))." %
+                    (point.x, point.y,
+                     imBBox.minX, imBBox.maxX, imBBox.minY, imBBox.maxY))
```
isullivan, line 233: I'm a little worried about this swallowing up other as-yet-unknown errors. Would it make sense to only return `None` for this case, and re-raise the error if it was a different problem?

isullivan (approved):
Looks good. My main question is whether we should catch all `InvalidParameterError`s, or only the ones that were caused by this particular bug. I assume we would want to catch them all, but we might want to hit them for now so that we can try to understand where they are coming from.

python/lsst/ap/association/packageAlerts.py
morriscb: It is an Extent2I. Do you want me to change the variable names?

isullivan: Yes please.

python/lsst/ap/association/packageAlerts.py
morriscb: So I looked over the code for cutouts in C++. The two errors this block can throw are both InvalidParameterError one for centroid outside of exposure bbox and the extent of the cutout bbox is <=0. There may be others further up stream but I haven't investigated that completely. I guess the best I can do here is restructure it to catch the error for outside centroid (by testing it) and then do a raise on anything else. Sound good?

PR #109, opened 2021-04-08
DM-29512: Investigate source centroids that fall off the image

morriscb:
Add analysis notebook.


mrawls (commented):
The notebook shows the work you did during this investigation; it's nothing fancy but it's fine to merge it if you think it will be a useful reference. Please consider adding a few more markdown cells to explain what the printed values, plots, etc. mean.

PR #110, opened 2021-04-20
DM-29147: Switch from using mapApData to new pipeline task and functors.

morriscb:


data/DiaSource.yaml
```diff
@@ -86,8 +93,42 @@ funcs:
     # trailLnL not implemented
     # trailChi2 not implemented
     # trailNdata not implemented
-    # dipMeanFlux needs functor DM-
-    # dipFluxDiff needs functor DM-
+    dipMeanFlux:
```
parejkoj, line 96: "dip" is really the name of it? Not `dipole`? That's very confusing to me: I saw "dip" and thought this was about a source whose flux had gone down or something.

parejkoj, line 96: Also, could you please put a note at the top of the file referencing or linking to where these fields are defined? They don't have docstrings in here, and I assume they are formally defined elsewhere. Not knowing the package, I don't know where to look that up.

data/DiaSource.yaml
```diff
@@ -131,23 +175,53 @@ funcs:
...
+            - base_LocalWcs_CDMatrix_1_2
+            - base_LocalWcs_CDMatrix_2_1
+            - base_LocalWcs_CDMatrix_2_2
     # Icov not implemented
     IxxPsf:
-        functor: Column
-        args: slot_PsfShape_xx
+        functor: ConvertPixelSqToArcsecondsSq
+        args:
+            - slot_Shape_xx
```
parejkoj, line 205: This looks like it would be identical in value to `Ixx`: Shouldn't the `IxxPsf` use an explicitly PSF-like shape? `slot_Shape_xx` is not necessarily PSF-like. I see you used `slot_PsfShape_xy` for `IxyPsf` below.

python/lsst/ap/association/diaPipe.py
```diff
@@ -50,22 +49,16 @@
...
-        name="{fakesType}{coaddName}Diff_diaSrc_schema",
-        storageClass="SourceCatalog",
-        multiple=True
-    )
-    diaSourceCat = connTypes.Input(
-        doc="Catalog of DiaSources produced during image differencing.",
-        name="{fakesType}{coaddName}Diff_diaSrc",
-        storageClass="SourceCatalog",
+    diaSourceTable = connTypes.Input(
+        doc=".",
```
parejkoj, line 59: That's not a doc.

python/lsst/ap/association/diaPipe.py
```diff
@@ -169,11 +162,6 @@ class DiaPipelineConfig(pipeBase.PipelineTaskConfig,
             "band not on this list, the appropriate band specific columns "
             "must be added to the Apdb schema in dax_apdb.",
     )
-    diaSourceDpddifier = pexConfig.ConfigurableField(
```
parejkoj, line 172: I just want to note that I am glad to see "Dpddifier" removed as a thing.

python/lsst/ap/association/transformDiaSourceCatalog.py
```diff
@@ -182,8 +183,11 @@ def run(self,
                             ParquetTable(dataFrame=diaSourceDf),
                             self.funcs,
                             dataId=None).df
+        # The Ra/DecColumn functors preserve the coord_ra/dec original columns.
+        # Since we don't need these and keeping them causes a DB insert crash
+        # we drop them from the DataFrame before returning out value.
```
parejkoj, line 188: "returning out value"?

python/lsst/ap/association/transformDiaSourceCatalog.py
```diff
@@ -182,8 +183,11 @@ def run(self,
...
         return pipeBase.Struct(
-            diaSourceTable=df
+            diaSourceTable=df.drop(columns=["coord_ra", "coord_dec"]),
```
parejkoj, line 190: Checking the docs, `df.drop()` defaults to returning the modified object. That results in a new copy of the data frame. I think what you want here is to make the `df.drop()` call before `return`, and pass `inplace=True` so that it modifies the dataframe in place without making a copy. See:

https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.drop.html

parejkoj (changes requested):
Some wording docs suggestions, but the `IxxPsf` and `df.drop()` questions are important ones, I think.

data/DiaSource.yaml
morriscb: Yep: http://ls.st/dpdd

morriscb: Added reference to http://ls.st/dpdd

data/DiaSource.yaml
morriscb: Thanks for the catch. Was that way due to a bad copy paste job. Now fixed.

python/lsst/ap/association/transformDiaSourceCatalog.py
morriscb: I understand that's what happening. What is your concern here of returning the copy of the dataframe that is computed in this method?

parejkoj: A general avoidance of creating unnecessary copies.

Though I've now spent some time on stackoverflow and some pandas docs, and it looks like an extra copy is typically created in this case even with `inplace=True` (there's a proposal to deprecate `inplace`: https://github.com/pandas-dev/pandas/issues/16529 ). This is surprising behavior, coming from the C++ side of the stack where "don't make unnecessary copies" was a big design goal.

So, I guess leave the code here as-is, and now I wonder whether we shouldn't mention this somewhere in the dev guide since we're starting to use pandas in more places?

data/DiaSource.yaml
parejkoj: That is very unfortunate, but not a problem for this ticket.

parejkoj (approved):

PR #111, opened 2021-05-03
DM-29812: Remove mapApData.py and tests.

morriscb:


isullivan (approved):
Looks good! I don't see anywhere else that `mapApData` is being used, so I think you can go ahead and remove it.

PR #112, opened 2021-05-04
DM 30015: Update conf.py for documenteer 0.6

jonathansick:
Update conf.py for the new documenteer 0.6 configuration API.  Based on
the latest updates in the package template
(https://github.com/lsst/templates/tree/master/project_templates/stack_package)

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -1,13 +1,13 @@
 """Sphinx configuration file for an LSST stack package.
-
-This configuration only affects single-package Sphinx documenation builds.
+This configuration only affects single-package Sphinx documentation builds.
+For more information, see:
+https://developer.lsst.io/stack/building-single-package-docs.html
 """
 
-from documenteer.sphinxconfig.stackconf import build_package_configs
-import lsst.ap.association
+from documenteer.conf.pipelinespkg import *
 
 
-_g = globals()
-_g.update(build_package_configs(
-    project_name='ap_association',
-    version=lsst.ap.association.version.__version__))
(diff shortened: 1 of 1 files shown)


timj:
@jonathansick do you want me to review some of these? They don't all have to be reviewed and merged on the same day. Incrementally works fine.

jonathansick:
If you'd like to review (and even merge) that'd be great. Thanks.

doc/conf.py
```diff
@@ -1,13 +1,13 @@
...
-
-This configuration only affects single-package Sphinx documenation builds.
+This configuration only affects single-package Sphinx documentation builds.
+For more information, see:
+https://developer.lsst.io/stack/building-single-package-docs.html
 """
 
-from documenteer.sphinxconfig.stackconf import build_package_configs
-import lsst.ap.association
+from documenteer.conf.pipelinespkg import *
```
timj, line 7: As a general comment (too late now given it's everywhere) it would seem less like magic and more PEP8 as:

```python
import documenteer.conf.pipelinespkg as p

p.project = "ap_association"
````

jonathansick: Did you check this with a Sphinx run? I'm not sure this would be compatible with the namespacing that Sphinx expects for configurations.

timj: Aha. I'll take a look. I'm sure you are right -- namespace shenanigans seems right.

timj: Right. Falls over completely. I'll merge this.

PR #113, opened 2021-05-12
DM-30030: Make calibration source selection consistent between gen2 and gen3 middleware - part deux

laurenam:


tests/test_loadDiaCatalogs.py
```diff
@@ -370,6 +370,7 @@ def testGetPixelRanges(self):
         the WCS.
         """
         diaConfig = LoadDiaCatalogsConfig()
+        diaConfig.pixelMargin = 300
```
TallJimbo, line 373: I presume this is necessary because the test has some sensitive thresholds you don't want to tinker with?  If so, please add a comment to that effect, so nobody thinks it's more important than that.

TallJimbo (approved):

PR #115, opened 2021-06-04
DM-30497: Bug: DiaSource table cannot find Ixx column

morriscb:


parejkoj:
Where can we put a consistency check for these values against those in `dax_apdb/data/apdb-schema.yaml`? This package depends on dax_apdb, so it's the obvious place for it. We need to validate that these column names match those in the official database schema.

morriscb:
Sure, it could work there. As far as I know there is supposed to eventually be a central place to test against for schemas derived from the DPDD. Colin would be the best person to ask about that I would guess.

parejkoj:
Why can't we test it here in this package?

tests/test_loadDiaCatalogs.py
```diff
@@ -295,6 +296,10 @@ def setUp(self):
         self.apdb.storeDiaObjects(self.diaObjects,
                                   self.dateTime)
 
+        # These columns intentionally appear in DiaSource.yaml,
+        # but not in the APDB schema
+        self.ignoreColumns = ["filterName", "bboxSize", "isDipole"]
```
morriscb, line 301: They are specified in the schema-extra file in ap_association.

tests/test_loadDiaCatalogs.py
```diff
@@ -390,6 +395,25 @@ def testGetPixelRanges(self):
             self.assertEqual(normPix, testPix)
             self.assertEqual(flipPix, testPix)
 
+    def test_apdbSchema(self):
```
morriscb, line 398: This only tests if the default schema for adpb and the functors run by default match. Also, it is safe to have extra columns in the apdb schema not filled by the DiaSource functor file. The main testing I've been doing to make sure things work has been running ap_verify as an integration test. That only just failed, as we've discovered, as sqlite is not case sensitive in it's column names but Postgres is. We can leave these here for now, but I will point out that it does not protect against mismatches in the schema files.

tests/test_loadDiaCatalogs.py
```diff
@@ -295,6 +296,10 @@ def setUp(self):
         self.apdb.storeDiaObjects(self.diaObjects,
                                   self.dateTime)
 
+        # These columns intentionally appear in DiaSource.yaml,
+        # but not in the APDB schema
```
parejkoj, line 300: Period at the end of sentence, please.

tests/test_loadDiaCatalogs.py
```diff
@@ -390,6 +395,25 @@ def testGetPixelRanges(self):
...
+        """Test that the DiaSource schema from dax_apdb agrees with the column
+        names defined here in ap_association/data/DiaSource.yaml.
+        """
+        functorFile = _data_file_name("DiaSource.yaml", "ap_association")
+        apdbSchemaFile = _data_file_name("apdb-schema.yaml", "dax_apdb")
+        with open(apdbSchemaFile) as yaml_stream:
+            table_list = list(yaml.safe_load_all(yaml_stream))
+            for table in table_list:
+                if table['table'] == 'DiaSource':
+                    apdbSchemaColumns = [column['name'] for column in table['columns']]
```
parejkoj, line 408: Wouldn't you then want a `break` after this, since there should only be one `DiaSource` table, so you no longer need to loop over the table_list?

tests/test_loadDiaCatalogs.py
```diff
@@ -390,6 +395,25 @@ def testGetPixelRanges(self):
...
+        with open(functorFile) as yaml_stream:
+            diaSourceFunctor = yaml.safe_load_all(yaml_stream)
+            for functor in diaSourceFunctor:
+                diaSourceColumns = list(functor['funcs'].keys())
+                for column in diaSourceColumns:
```
parejkoj, line 413: Instead of this loop, could you do `self.assertEqual(set(diaSourceColumns), set(apdbSchemaColumns))`, or maybe `self.assertLess(set(diaSourceColumns), set(apdbSchemaColumns))` since I think the one is a proper subset of the other?

tests/test_loadDiaCatalogs.py
parejkoj: How does this not protect against mismatches in the schema files? This test ensures that everything is case-consistent as well.

parejkoj (commented):
Thanks for adding the test; I have some suggestions about reworking it.

I would prefer if this could be a test against an actually instantiated database, instead of just testing yaml file contents, but that might be too tricky to implement.

tests/test_loadDiaCatalogs.py
morriscb: The schemas can be set at run time. This only tests for the default ones. It doesn't even pull in all columns that are currently being used and written to the apdb.

tests/test_loadDiaCatalogs.py
mrawls: Thanks, I figured there was someplace these not-DPDD-yet-important columns lived.

mrawls: Actually, I see `filterName` and `isDipole` in there, but where does `bboxSize` come from?

tests/test_loadDiaCatalogs.py
```diff
@@ -390,6 +395,25 @@ def testGetPixelRanges(self):
...
+            for table in table_list:
+                if table['table'] == 'DiaSource':
+                    apdbSchemaColumns = [column['name'] for column in table['columns']]
+                    break
+        with open(functorFile) as yaml_stream:
+            diaSourceFunctor = yaml.safe_load_all(yaml_stream)
+            for functor in diaSourceFunctor:
+                diaSourceColumns = [column for column in list(functor['funcs'].keys())
+                                    if column not in self.ignoreColumns]
+                self.assertLess(set(diaSourceColumns), set(apdbSchemaColumns))
```
parejkoj, line 415: This assert should go outside the `for` loop: you want to build up the list of column names from the `DiaSource.yaml` file, and then check against that whole list.

PR #116, opened 2021-06-16
DM-30386: Make DRP wrapper task for DiaCalculation task

morriscb:


python/lsst/ap/association/association.py
```diff
@@ -167,7 +167,7 @@ def run(self,
             diaObjects,
             mergedDiaSourceHistory,
             matchResult.associated_dia_object_ids,
-            filterName)
+            [filterName])
```
kherner, line 170: One question here: how are the filter name(s) passed? They're not already a list, are they? If I look at e.g. python/lsst/pipe/tasks/drpDiaCalculationPipe.py you have 

self.config.filterNames

passed, so it's read from the config. In the actual config is it something like "u","g","r",... as opposed to ["u", "g", ...] my only concern was that you might unexpectedly get a list of lists somehow.

morriscb: filterName here comes from the diaSource catalog itself and will always be a string value here. If you look a few lines above this, it is where filterName is pulled. The pipeline here will only ever be run over one filter as it is designed to run in AP that is one ccdVisit at time. This one runs in the DiaPipeTask pipeline task vs the new wrapper.

kherner (commented):
One question on the deleted diaCalculation.py and diaCalculationPlugins.py files. When I compare the lines in the "new" versions in meas_base I see there are 514 and 1,110 lines, respectively, but 463 and 1,093 in the now-deleted versions. The test files are the same though. Do you know off the top of your head where the difference come from? I cna do a detailed comparison but if this expected maybe we can save some time.

morriscb:
It's due to 1) the extra for loop over plugins outside of the one that loops over filters first. 2) adding the `needsFilter` boolean to each plugin.

PR #117, opened 2021-06-17
DM-30798: Update to new adjustQuantum signature.

TallJimbo:
This is a belated fix for an API change on DM-30649.

### DIFF
--- a/python/lsst/ap/association/diaPipe.py
+++ b/python/lsst/ap/association/diaPipe.py
@@ -102,7 +102,7 @@ class DiaPipelineConnections(
         if not config.doWriteAssociatedSources:
             self.outputs.remove("associatedDiaSources")
 
-    def adjustQuantum(self, datasetRefMap: pipeBase.InputQuantizedConnection):
+    def adjustQuantum(self, inputs, outputs, label, dataId):
         """Override to make adjustments to `lsst.daf.butler.DatasetRef` objects
         in the `lsst.daf.butler.core.Quantum` during the graph generation stage
         of the activator.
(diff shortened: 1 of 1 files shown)


python/lsst/ap/association/diaPipe.py
```diff
@@ -112,23 +112,53 @@ def adjustQuantum(self, datasetRefMap: pipeBase.InputQuantizedConnection):
...
+            connection name and whose values are a tuple of the connection
+            instance and a collection of associated `DatasetRef` objects.
+            The exact type of the nested collections is unspecified; it can be
+            assumed to be multi-pass iterable and support `len` and ``in``, but
+            it should not be mutated in place.  In contrast, the outer
+            dictionaries are guaranteed to be temporary copies that are true
+            `dict` instances, and hence may be modified and even returned; this
+            is especially useful for delegating to `super` (see notes below).
+        outputs : `Mapping`
+            Mapping of output datasets, with the same structure as ``inputs``.
```
kfindeisen, line 126: Did you intend for `inputs` to be `dict` but `outputs` (and everything else) to be `Mapping`?

kfindeisen (approved):
One question about the documentation.

python/lsst/ap/association/diaPipe.py
TallJimbo: Oh, good catch. I did not, and that's a mistake down in `pipe_base`, too (should be `dict` throughout).  I'll fix it here on this branch, but fix up pipe_base on a different ticket in that package that I already have underway.

PR #119, opened 2021-07-02
DM-30985: Use lazy evaluation of functor file and flag map

timj:
Using getPackageDir means that the path is burned into the config on read but this leads to the config stored in quantum graphs having a hard coded path to the location of ap_association at that time. For some workflow systems the path to the software on the execution node is in a different place so this breaks everything. This change embeds the environment variable in the config and relies on the runtime to call expandvars.

### DIFF
--- a/python/lsst/ap/association/transformDiaSourceCatalog.py
+++ b/python/lsst/ap/association/transformDiaSourceCatalog.py
@@ -82,7 +82,7 @@ class TransformDiaSourceCatalogConfig(pipeBase.PipelineTaskConfig,
         dtype=str,
         doc='Path to YAML file specifying Science DataModel functors to use '
             'when copying columns and computing calibrated values.',
-        default=os.path.join(getPackageDir("ap_association"),
+        default=os.path.join("${AP_ASSOCIATION_DIR}",
                              "data",
                              "DiaSource.yaml")
     )
--- a/python/lsst/ap/association/transformDiaSourceCatalog.py
+++ b/python/lsst/ap/association/transformDiaSourceCatalog.py
@@ -34,7 +34,6 @@ import lsst.pipe.base as pipeBase
 import lsst.pipe.base.connectionTypes as connTypes
 from lsst.pipe.tasks.postprocess import TransformCatalogBaseTask
 from lsst.pipe.tasks.parquetTable import ParquetTable
-from lsst.utils import getPackageDir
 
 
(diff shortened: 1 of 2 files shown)


python/lsst/ap/association/transformDiaSourceCatalog.py
```diff
@@ -82,7 +82,7 @@ class TransformDiaSourceCatalogConfig(pipeBase.PipelineTaskConfig,
         dtype=str,
         doc='Path to YAML file specifying Science DataModel functors to use '
             'when copying columns and computing calibrated values.',
-        default=os.path.join(getPackageDir("ap_association"),
+        default=os.path.join("${AP_ASSOCIATION_DIR}",
```
kfindeisen, line 84: Is this a change we should be making in all configs? I think I've used `getPackageDir` _everywhere_ (including `ap_verify` data sets).

kfindeisen (approved):
Looks good. Thanks for the PR description, otherwise I would have had no idea what was going on.

python/lsst/ap/association/transformDiaSourceCatalog.py
timj: I made a comment in pipelines slack about this yesterday. getPackageDir is fine if you are using it to find other configuration information. The problem comes when the getPackageDir value ends up being assigned to a config item which is then persisted in a quantum graph.  @TallJimbo made the suggestion that we should expand ButlerURI (after extracting it from daf_butler) to have a URI scheme that could be used for these cases (e.g., `eups://ap_association/data/blah.yaml`)  but that's a longer term story and I needed something to work quickly because DP0.2 PanDA testing can't work at the moment. Personally I think we should start using python pkg_resources (which I made daf_butler use to find its information since getPackageDir wasn't an option) but I haven't got much traction on that. But yes, if a path is burned into a config then lazy evaluation is preferred.

kfindeisen: Ah. In that case, I think you missed a spot: https://github.com/lsst/ap_association/blob/15c94b941f83be4f9b625bbae886c9ec6bba7fd6/python/lsst/ap/association/diaPipe.py#L228.

timj: Thanks. I've just pushed a fix for this one and that opened up the can of worms in dax_apdb resulting in lsst/dax_apdb#19 -- I'm happy for you to take a look at those changes as well but I have also tagged @andy-slac.

This will definitely all be improved by using URI schemes.

PR #121, opened 2021-07-23
DM-29346: Compute PSFlux chi squared metric

parejkoj:
Have to skip these columns until we update apdb on DM-31193.


kfindeisen (approved):
I think this is fine, though I don't really understand the distinction between `base_*` and `slot_*`.

parejkoj:
The `base_` vs. `slot_` duplications in `dropColumns` can probably be fixed by using `schema.disconnectAliases()` before doing the pandas conversion to remove the slot aliases. That will require further work, however, since the `slot_` fields are used in the existing code.

PR #128, opened 2021-09-16
DM-31313: Add pixelId calculation

yalsayyad:
Fix formating.


yalsayyad (commented):
Apparently I can't hit the approve button if I made the PR.  Reviewed!

PR #131, opened 2021-09-28
DM-31825: Implement RFC-807 on DRP parquet table consistency

yalsayyad:
-  Adapt to parent class optional primaryKey config.
-  Inherit config from base task
-  Add config parameter to switch whether flags are transformed as one bit-packed column named 'flags' or individual boolean columns  with column rename rules applied.


morriscb (approved):
Looks good.

PR #132, opened 2021-10-01
DM-31619: Fully incorporate SSP object association in DiaPipe

morriscb:


python/lsst/ap/association/diaPipe.py
```diff
@@ -202,6 +211,15 @@ class DiaPipelineConfig(pipeBase.PipelineTaskConfig,
         target=AssociationTask,
         doc="Task used to associate DiaSources with DiaObjects.",
     )
+    doSolarSystemAssociation = pexConfig.Field(
+        dtype=bool,
+        default=False,
+        doc="",
```
ebellm, line 217: While I think the name is pretty self-explanatory here please do give this a non-null docstring.

python/lsst/ap/association/diaPipe.py
```diff
@@ -469,21 +506,22 @@ def createNewDiaObjects(self, diaSources):
...
         newDiaObjectsList = []
-        for idx, diaSource in diaSources.iterrows():
-            if diaSource["diaObjectId"] == 0:
-                newDiaObjectsList.append(
-                    self._initialize_dia_object(diaSource["diaSourceId"]))
-                diaSources.loc[idx, "diaObjectId"] = diaSource["diaSourceId"]
+        for idx, diaSource in unAssocDiaSources.iterrows():
+            newDiaObjectsList.append(
+                self._initialize_dia_object(diaSource["diaSourceId"]))
+            unAssocDiaSources.loc[idx, "diaObjectId"] = diaSource["diaSourceId"]
```
ebellm, line 515: I'm confused here as to why you are setting the `diaObjectId` to be the `diaSourceId`.  Are these not distinct ids?

tests/test_diaPipe.py
```diff
@@ -52,28 +55,41 @@ def setUp(self):
     def tearDown(self):
         pass
 
+    def testRun(self):
+        """Test running while creating and packaging alerts.
+        """
+        self._testRun(True, True)
```
ebellm, line 61:
```suggestion
        self._testRun(doPackageAlerts=True, doSolarSystemAssociation=True)
```

tests/test_diaPipe.py
```diff
@@ -52,28 +55,41 @@ def setUp(self):
...
+
+    def testRunWithSolarSystemAssociation(self):
+        """Test running while creating and packaging alerts.
+        """
+        self._testRun(False, True)
```
ebellm, line 66:
```suggestion
        self._testRun(doPackageAlerts=False, doSolarSystemAssociation=True)
```

tests/test_diaPipe.py
```diff
@@ -52,28 +55,41 @@ def setUp(self):
...
+
     def testRunWithAlerts(self):
         """Test running while creating and packaging alerts.
         """
-        self._testRun(True)
+        self._testRun(True, False)
```
ebellm, line 71:
```suggestion
        self._testRun(doPackageAlerts=True, doSolarSystemAssociation=False)
```

tests/test_diaPipe.py
```diff
@@ -52,28 +55,41 @@ def setUp(self):
...
 
-    def testRunWithoutAlerts(self):
+    def testRunWithoutAlertsOrSolarSystem(self):
         """Test running without creating and packaging alerts.
         """
-        self._testRun(False)
+        self._testRun(False, False)
```
ebellm, line 76:
```suggestion
        self._testRun(doPackageAlerts=False, doSolarSystemAssociation=False)
```

python/lsst/ap/association/diaPipe.py
```diff
@@ -469,21 +506,22 @@ def createNewDiaObjects(self, diaSources):
...
         """
         newDiaObjectsList = []
-        for idx, diaSource in diaSources.iterrows():
-            if diaSource["diaObjectId"] == 0:
-                newDiaObjectsList.append(
-                    self._initialize_dia_object(diaSource["diaSourceId"]))
-                diaSources.loc[idx, "diaObjectId"] = diaSource["diaSourceId"]
+        for idx, diaSource in unAssocDiaSources.iterrows():
+            newDiaObjectsList.append(
+                self._initialize_dia_object(diaSource["diaSourceId"]))
```
ebellm, line 514: I would investigate if you can run this without the `iterrows` loop, for example:

`newDiaObjectsSeries = unAssocDiaSources['diaSourceId'].apply(self._initialize_dia_object)`

which if it works will likely be faster than the explicit loop.

python/lsst/ap/association/diaPipe.py
ebellm: In any case this doesn't need to be looped: you can simply say 

`unAssocDiaSources['diaObjectId'] = unAssocDiaSources['diaSourceId']`

ebellm (changes requested):
A few comments and suggestions throughout.

python/lsst/ap/association/diaPipe.py
morriscb: The id for newly created DiaObjects has been the same as the id of the first DiaSource that makes up the object since we've had DiaObjects in the pipeline. It's an easy way to give all the DiaObjects a unique id.

python/lsst/ap/association/diaPipe.py
morriscb: I'll give it a shot.

morriscb: Seems to work. I remember apply being only slightly faster than iterrows when not using built in pandas operations. Still, it's likely to be a bit faster since it now operates on one column not the whole row.

PR #142, opened 2021-12-02
DM-32682: Mock out metadata setting in test

timj:
With the task mocked out the metadata values calculated
by the task are no longer real values but are themselves
MagicMock objects. MagicMock objects can't be stored in
a TaskMetadata so we need to mock out the API that
sets the metadata.

This works with PropertySet since that seems to accept
the value and silently ignores it.

### DIFF
--- a/tests/test_diaPipe.py
+++ b/tests/test_diaPipe.py
@@ -25,8 +25,9 @@ import pandas as pd
 
 import lsst.afw.image as afwImage
 import lsst.afw.table as afwTable
-import lsst.pipe.base as pipeBase
+from lsst.pipe.base.testUtils import assertValidOutput
 import lsst.utils.tests
+import lsst.utils.timer
 from unittest.mock import patch, Mock, MagicMock, DEFAULT
 
 from lsst.ap.association import DiaPipelineTask
@@ -96,7 +97,6 @@ class TestDiaPipelineTask(unittest.TestCase):
         # appropriately.
         subtasksToMock = [
             "diaCatalogLoader",
-            "associator",
             "diaCalculation",
             "diaForcedSource",
         ]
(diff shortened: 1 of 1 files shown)


tests/test_diaPipe.py
```diff
@@ -114,21 +114,31 @@ def _testRun(self, doPackageAlerts=False, doSolarSystemAssociation=False):
         # execution in the test environment.
```
ktlim, line 114: This comment should move down to just before `with patch.multiple(`.

tests/test_diaPipe.py
```diff
@@ -114,21 +114,31 @@ def _testRun(self, doPackageAlerts=False, doSolarSystemAssociation=False):
...
-                                  diffIm,
-                                  exposure,
-                                  template,
-                                  ccdExposureIdBits,
-                                  "g")
-                for subtaskName in subtasksToMock:
-                    getattr(task, subtaskName).run.assert_called_once()
-                pipeBase.testUtils.assertValidOutput(task, result)
+                with patch('lsst.ap.association.diaPipe.DiaPipelineTask._add_association_meta_data',
+                           new=metadataMock):
```
ktlim, line 131: I guess I'd prefer that the subtask mocks actually return numbers that could be inserted into the metadata.  Is mocking things this way overly constraining the interface of the internal function, requiring changes here if additional items are added?

ktlim (approved):

tests/test_diaPipe.py
timj: Okay. I'll try to work out how the mocks work. This was the quick fix because I didn't want to disentangle all the mocking.

ktlim: That isn't necessarily your job; you can merge this and file a ticket for the package author(s) to fix this.

timj:
@ktlim  I took another look and realized that if I mocked out the `run()` method of the subtask and not the entire subtask, then I could make the test store valid metadata.

tests/test_diaPipe.py
```diff
@@ -105,20 +105,37 @@ def _testRun(self, doPackageAlerts=False, doSolarSystemAssociation=False):
...
+                                         unAssocDiaSources=MagicMock(spec=pd.DataFrame()))
+
+        # apdb isn't a subtask, but still needs to be mocked out for correct
+        # execution in the test environment.
         with patch.multiple(
             task, **{task: DEFAULT for task in subtasksToMock + ["apdb"]}
         ):
-            with patch('lsst.ap.association.diaPipe.pd.concat',
-                       new=concatMock):
+            with patch('lsst.ap.association.diaPipe.pd.concat', new=concatMock), \
```
ktlim, line 134: I believe you can parenthesize this whole thing, eliminating the need for backslashes.  And can you add in the `patch.multiple()` as well?

ktlim (approved):

tests/test_diaPipe.py
ktlim: Oh, sorry, the parenthesized list may be Python 3.9.

ktlim: Actually Python 3.10.

timj: I didn't merge it with the `patch.multiple` because I didn't want to change the indent level of the code I wasn't touching.

PR #145, opened 2022-04-07
DM-34363: Add trailFlux measurement to ap_association

langfzac:
Adds `trailFlux` to `DiaSources.yaml`.

Note: Flux error is not implemented yet, so a place-holder is given to the functor. That argument is not used in the transformation anyways.

### DIFF
--- a/data/DiaSource.yaml
+++ b/data/DiaSource.yaml
@@ -90,7 +90,13 @@ funcs:
     psNdata:
         functor: Column
         args: slot_PsfFlux_npixels
-    # trailFlux not implemented
+    trailFlux:
+        functor: LocalNanojansky
+        args:
+            - ext_trailedSources_Naive_flux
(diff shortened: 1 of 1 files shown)


data/DiaSource.yaml
```diff
@@ -90,7 +90,13 @@ funcs:
     psNdata:
         functor: Column
         args: slot_PsfFlux_npixels
-    # trailFlux not implemented
+    trailFlux:
+        functor: LocalNanojansky
+        args:
+            - ext_trailedSources_Naive_flux
+            - slot_ApFlux_instFluxErr  # Place holder; arg not actually used
```
kfindeisen, line 97: So should this eventually be `ext_trailedSource_Naive_fluxerr`?

kfindeisen (approved):
Looks good!

data/DiaSource.yaml
langfzac: Yes, once that's implemented.

PR #147, opened 2022-05-04
DM-32694: Split AP pipeline into ApPipeWithFakes

kfindeisen:
This PR changes `DiaPipelineTask` to use `fakes_calexp` when appropriate, consistent with other input datasets.


mrawls (approved):
Good catch

PR #148, opened 2022-05-05
DM-33821: Update docstrings for solar system metrics.

kfindeisen:
This PR fixes some copy-pasted docstrings, borrowing text from the code that calculated the metadata in the first place. It also documents the ephemeris catalog format so that we know what information we have for future development.


isullivan (approved):
These updates look fine.

PR #149, opened 2022-05-18
DM-31934: Replace HH:MM:SS conversion code with astropy equivalents.

ebellm:


python/lsst/ap/association/skyBotEphemerisQuery.py
```diff
@@ -231,10 +233,12 @@ def _skybotConeSearch(self, expCenter, epochJD, queryRadius):
...
             coldict = dict(zip(dfSSO.columns, columns))
             dfSSO.rename(columns=coldict, inplace=True)
-            # TODO: DM-31934 Replace custom conversion code with
-            #       Astropy.coordinates.
-            dfSSO["ra"] = self._rahms2radeg(dfSSO["RA(h)"])
-            dfSSO["decl"] = self._decdms2decdeg(dfSSO["DE(deg)"])
+            # Data returned in hourangle format.
+            radec = dfSSO["RA(h)"]+' '+dfSSO["DE(deg)"]
+            skycoo = SkyCoord(radec.values, unit=(u.hourangle, u.deg))
+            # Use astropy SkyCoord to convert to decimal degrees.
```
kfindeisen, line 239: This comment would make more sense before the line that uses `SkyCoord`.

kfindeisen (approved):
Looks good, style comments only.

python/lsst/ap/association/skyBotEphemerisQuery.py
```diff
@@ -231,10 +233,12 @@ def _skybotConeSearch(self, expCenter, epochJD, queryRadius):
             columns = [col.strip() for col in dfSSO.columns]
             coldict = dict(zip(dfSSO.columns, columns))
             dfSSO.rename(columns=coldict, inplace=True)
-            # TODO: DM-31934 Replace custom conversion code with
-            #       Astropy.coordinates.
-            dfSSO["ra"] = self._rahms2radeg(dfSSO["RA(h)"])
-            dfSSO["decl"] = self._decdms2decdeg(dfSSO["DE(deg)"])
+            # Data returned in hourangle format.
+            radec = dfSSO["RA(h)"]+' '+dfSSO["DE(deg)"]
+            skycoo = SkyCoord(radec.values, unit=(u.hourangle, u.deg))
```
timj, line 238: Isn't it a bit overkill to use `SkyCoord` here?

```python
from astropy.coordinates import Angle

dfSSO["ra"] = Angle(dfSSO["RA(h)"])
dfSSO["decl"] = Angle(dfSSO["DE(deg)", unit="deg")
```
eg.:
```python
>>> c.Angle("5 32 3.4", unit="deg").deg
5.534277777777778
>>> c.Angle("5 32 3.4", unit="h").deg
83.01416666666665
```

kfindeisen: I would have used `Angle` myself, yes, but the most important thing is to have an off-the-shelf solution.

timj: SkyCoord adds an extra layer of indirection and requires you to concatenate the two strings. I agree that anything is better than doing it ourselves.

ebellm: The optimization here seems pretty tiny--we're going to throw this whole module away--but sure, that works.

PR #153, opened 2022-06-24
DM-34801: Replace Diff_warpedExp with Diff_templateExp in diaPipe

parejkoj:


python/lsst/ap/association/diaPipe.py
```diff
@@ -81,12 +81,12 @@ class DiaPipelineConnections(
...
         dimensions=("instrument", "visit", "detector"),
     )
-    warpedExposure = connTypes.Input(
+    template = connTypes.Input(
         doc="Warped template used to create `subtractedExposure`. Not PSF "
             "matched.",
         dimensions=("instrument", "visit", "detector"),
         storageClass="ExposureF",
-        name="{fakesType}{coaddName}Diff_warpedExp",
+        name="{fakesType}{coaddName}Diff_templateExp",
```
kfindeisen, line 89: I think we're using the default in all the AP pipelines. Will this change break them?

kfindeisen (approved):
Looks good!

python/lsst/ap/association/diaPipe.py
parejkoj: I'm testing this branch by running ap_verify, which should be enough? I think I'll have to run with both the `ApVerify.yaml` and the `ApVerifyWithFakes.yaml` to be sure.

kfindeisen: Yes, it should be enough, unless there are any old-style config overrides lurking about.

FTR, `ap_verify` is _weird_, so I'd recommend testing with `pipetask run` instead. If you use `ingest_datasets.py` to set up the dataset, you can grab appropriately configured pipelines from within the workspace it creates.

parejkoj: Yes, I've just verified that this works with both `ApVerify.yaml` and `ApVerifyWithFakes.yaml`.

PR #154, opened 2022-06-24
DM-34539: Add detection significance to DIASources and apdb

parejkoj:


python/lsst/ap/association/transformDiaSourceCatalog.py
```diff
@@ -202,6 +208,22 @@ def run(self,
         if self.config.doRemoveSkySources:
             diaSourceDf = diaSourceDf[~diaSourceDf["sky_source"]]
         diaSourceDf["bboxSize"] = self.computeBBoxSizes(diaSourceCat)
+
+        def getSignificance():
+            """Return the significance value of the first peak in each source
+            footprint."""
+            size = len(diaSourceDf)
```
ebellm, line 215: This has a bug--if `self.config.doRemoveSkySources==True`, then `len(diaSourceDf) != len(diaSourceCat)` since you removed the sky sources from the dataframe in line 209 above.

If you use `size = len(diaSourceCat)` I think it will work correctly.

ebellm (changes requested):

python/lsst/ap/association/transformDiaSourceCatalog.py
parejkoj: I believe that this is actually correct as written: it runs in ap_verify_ci_hits2015, whereas your version fails with:

```
ValueError: Length of values (274) does not match length of index (174)
```

Four lines down from this, I check whether a sources is a sky source and skip it if we remove sky sources, which is what allows this to work. It's not an ideal approach, I agree.

Note that I filed DM-35331 about this: we don't actually test the skySource code branches in this Task, except in ap_verify.

ebellm: No, I'm certain it's a bug.  You're correct that the solution I proposed above doesn't work, though.

This only matters if we remove sky sources.  Say there are 200 DIASources, of which 100 are sky sources.  Your line 209 will make `diaSourceDf` have `len` 100.  So line 217 will iterate over `size=100` records--but `diaSourceCat` has 200 records, so you will only look at the first 100 of them--skipping the sky sources doesn't give you extra iterations.  You'll end up with a `result` that has 100 elements which are a mix of real SNRs and nans.

I think the easiest workaround is just to do the sky source filtering after the `getSignificance` call in line 226.  See https://github.com/lsst/ap_association/tree/u/ebellm/DM-34539

parejkoj: Ok, yes, I see it now. Your re-ordering of the internals there definitely makes me want to tackle DM-35331 ASAP. I think `computeBBoxSizes` is actually fine, but that all seems pretty fragile.

parejkoj: I've incorporated your fix (squashed into my commit) and confirmed it works in ap_verify ci_cosmos. Thanks for catching that!

ebellm (approved):

PR #155, opened 2022-07-21
DM-35331: Add tests of sky source handling in transformDiaSourceCat

parejkoj:


tests/test_transformDiaSourceCatalog.py
```diff
@@ -41,6 +41,7 @@
 class TestTransformDiaSourceCatalogTask(unittest.TestCase):
     def setUp(self):
         self.nSources = 10
+        # Default PSF size for TestDataset gives an 18 pixel source box.
         self.bboxSize = 18
```
kfindeisen, line 46: Can you force the sizes to match, instead of relying on defaults?

tests/test_transformDiaSourceCatalog.py
```diff
@@ -105,7 +105,8 @@ def test_run(self):
         # The final snr value should be NaN because it doesn't have a peak significance field.
         expect_snr = [397.887353515625]*9
         expect_snr.append(np.nan)
-        self.assertTrue(np.array_equal(result.diaSourceTable["snr"], expect_snr, equal_nan=True))
+        # Have to use allclose because assert_array_equal doesn't support equal_nan.
+        np.testing.assert_allclose(result.diaSourceTable["snr"], expect_snr, equal_nan=True, rtol=0)
```
kfindeisen, line 114: I'm not sure the comment is necessary -- since the transformation seems to be doing unit or convention conversions (see comments above about significance vs. SNR), isn't it unsafe to assume exact equality anyway?

tests/test_transformDiaSourceCatalog.py
```diff
@@ -102,12 +106,36 @@ def test_run(self):
         np.testing.assert_array_equal(result.diaSourceTable["midPointTai"],
                                       [self.date.get(system=dafBase.DateTime.MJD)]*self.nSources)
         np.testing.assert_array_equal(result.diaSourceTable["diaObjectId"], [0]*self.nSources)
+        np.testing.assert_array_equal(result.diaSourceTable["x"], np.arange(self.nSources))
```
kfindeisen, line 109: I don't understand what the new `x` and `y` fields have to do with testing sky sources?

python/lsst/ap/association/transformDiaSourceCatalog.py
```diff
@@ -204,34 +204,29 @@ def run(self,
...
         diaSourceDf["bboxSize"] = self.computeBBoxSizes(diaSourceCat)
         diaSourceDf["ccdVisitId"] = ccdVisitId
         diaSourceDf["filterName"] = band
         diaSourceDf["midPointTai"] = diffIm.getInfo().getVisitInfo().getDate().get(system=DateTime.MJD)
         diaSourceDf["diaObjectId"] = 0
         diaSourceDf["ssObjectId"] = 0
 
+        if self.config.doRemoveSkySources:
+            diaSourceDf = diaSourceDf[~diaSourceDf["sky_source"]]
+
```
kfindeisen, line 229: Why change the  order here? It seems a bit silly to do the column updates before trimming out the rows.

kfindeisen (approved):
Looks good, minor comments only.

tests/test_transformDiaSourceCatalog.py
parejkoj: Unfortunately, it's coming from the default `psfDim` kwarg to `makeEmptyExposure` in TestDataset. I could tie this to the psf bounding box +1, but that doesn't feel right (and I don't think it's always +1; that probably depends on where within the pixel the source is).

tests/test_transformDiaSourceCatalog.py
parejkoj: I believe in this case exact equality is what we want: there should be no numeric change in this value between the source catalog and the pandas table.

tests/test_transformDiaSourceCatalog.py
parejkoj: I wanted another way to distinguish that the source ordering was correct: I did this ticket because in my last ticket on this package I got the order wrong in my first attempt, and this test would have caught that.

tests/test_transformDiaSourceCatalog.py
kfindeisen: Could you read the box size from the test dataset once it's created?

python/lsst/ap/association/transformDiaSourceCatalog.py
parejkoj: I agree that updating all the columns and then removing some rows isn't ideal, but it made the code (especially `getSignificance` and `computeBBoxSizes`) a lot clearer, I think.

Having the sky source removal happen in the middle of the source assignment (`diaSourceDf["snr"] = getSignificance()` needs to happen before sky sources are removed, and the previous version of `computeBBoxSizes()` had to happen after) made it more fragile: it wasn't immediately obvious where to insert a new computed field. Doing all the sky source removal at once at the end was much simpler than having extra logic inside those two methods to determine whether to process a sky source or not. I don't think has any significant effect on the runtime.

tests/test_transformDiaSourceCatalog.py
parejkoj: `exposure.psf.computeBBox()` gives 17: I'm guessing there's code within ap_association that ensures the source boxes are even? It's tied to the default for `TestDataset`, and I doubt that would ever be changed.

parejkoj: I've tweaked the comment a bit, though.

tests/test_transformDiaSourceCatalog.py
parejkoj: What do you mean by "see comments above about significance vs. SNR"? I didn't see a comment from you related to those two.

kfindeisen: I meant the pre-existing comments in the code. 🙂 

But I'm not very familiar with what transform actually does, so if you say there are no conversions, I'll take your word for it.

python/lsst/ap/association/transformDiaSourceCatalog.py
parejkoj: It turns out that you can do boolean slicing on afw catalogs! I've pulled out `getSignificance` to a free function and I do the filtering first now. I think that is an improvement.

PR #156, opened 2022-08-04
DM-35331-hotfix: Make outputBBoxSizes an array of dtype int

arunkannawadi:


erykoff (approved):

arunkannawadi:
Will merge after Jenkins passes with `ci_imsim` and `ci_hsc`

PR #157, opened 2022-08-04
DM-35331: Add note about int requirement, and restore np.empty

parejkoj:
I'm restoring the use of `empty` here because that was an intentional choice (all the fields will get initialized in the loop), and it's good to have a note as to why that `int` really is necessary.


arunkannawadi (approved):
Good to go once Jenkins passes with ci_imsim

parejkoj:
Jenkins with ci_imsim: https://ci.lsst.codes/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/37143/pipeline

PR #162, opened 2022-10-19
DM-36058: Fix untested Pandas deprecation warnings in ap_association

kfindeisen:
This PR removes references to `Dataset.append` that weren't covered by unit tests, and tweaks the tests so that they don't make assumptions about the call signatures in application code. It does not try to expand test coverage to perform actual Pandas operations, since I don't see a way to introduce mock tables without making them brittle to external changes.

### DIFF
--- a/python/lsst/ap/association/diaPipe.py
+++ b/python/lsst/ap/association/diaPipe.py
@@ -383,8 +383,9 @@ class DiaPipelineTask(pipeBase.PipelineTask):
                                        inplace=True)
 
         # Append new DiaObjects and DiaSources to their previous history.
-        diaObjects = loaderResult.diaObjects.append(
-            createResults.newDiaObjects.set_index("diaObjectId", drop=False),
+        diaObjects = pd.concat(
+            [loaderResult.diaObjects,
+             createResults.newDiaObjects.set_index("diaObjectId", drop=False)],
             sort=True)
         if self.testDataFrameIndex(diaObjects):
             raise RuntimeError(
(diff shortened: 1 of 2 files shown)


python/lsst/ap/association/diaPipe.py
```diff
@@ -383,8 +383,9 @@ def run(self,
                                        inplace=True)
 
         # Append new DiaObjects and DiaSources to their previous history.
-        diaObjects = loaderResult.diaObjects.append(
-            createResults.newDiaObjects.set_index("diaObjectId", drop=False),
+        diaObjects = pd.concat(
```
parejkoj, line 386: I'm very worried about using concat in our code like this: it can change the schema out from under us. Should we be using [convert_dtypes](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.convert_dtypes.html) here (and anywhere we use concat)? Could we move this away from pandas entirely and use `numpy.concatenate` or `numpy.vstack`?

If we had a way to get `pd.concat` to `raise` instead of silently casting to float, that might be more reasonable, but I don't think there is.

parejkoj (commented):
My only comment is my growing concern about `pd.concat`. I don't know what to do about this, though, if we're going to continue to use pandas in diaPipe.

parejkoj:
See discussion on #160 : I've just found that the diaForcedSource `append` is currently changing the dtype of the `flags` field. So, possibly your change here doesn't actually alter how that particular bug behaves, but we should definitely audit our use of `pd.concat` and replace it with something that doesn't cast everything to `float`.

That's for a different ticket (not filed yet), so I think you're good to merge here.

python/lsst/ap/association/diaPipe.py
kfindeisen: I don't know anything about `convert_dtypes`, I just followed the instructions for how to phase out `append`. As for using pandas in general, see [my comment](https://github.com/lsst/ap_association/pull/160#issuecomment-1290868299) on #160.

PR #165, opened 2023-01-20
DM-37622: changes for running with HSM shape

kherner:


data/association-flag-map.yaml
```diff
@@ -79,18 +79,18 @@ columns:
...
-    - name: slot_Shape_flag_shift
-      bit: 26
-      doc: centroid shifted by more than the maximum allowed amount
-    - name: slot_Shape_flag_maxIter
-      bit: 27
-      doc: Too many iterations in adaptive moments
-    - name: slot_Shape_flag_psf
-      bit: 28
-      doc: Failure in measuring PSF model shape
+#    - name: slot_Shape_flag_unweightedBad
```
parejkoj, line 82: Please add a comment above this block about why all of these were commented out.

parejkoj (commented):
Not having `slot_Shape_flag_psf` is a bit worrisome: Does HsmShape set differently-named flags with similar meanings?

data/association-flag-map.yaml
kherner: I've removed all of the SdssShape values instead of commenting them out, and replaced them with the flags from HsmShape.cc.

kfindeisen (commented):
I'm a bit worried that this change breaks compatibility with the old association flags. Are we fine with this change? Or is it just that nobody used the flag before, so it doesn't matter?

data/association-flag-map.yaml
```diff
@@ -79,18 +79,15 @@ columns:
...
-      doc: centroid shifted by more than the maximum allowed amount
-    - name: slot_Shape_flag_maxIter
+      doc: parent source, ignored
+    - name: slot_Shape_flag_galsim
       bit: 27
-      doc: Too many iterations in adaptive moments
-    - name: slot_Shape_flag_psf
-      bit: 28
-      doc: Failure in measuring PSF model shape
+      doc: GalSim failure
```
kfindeisen, line 93: Sorry, missed this the first time around -- why are we using GalSim in the AP pipeline in the first place?

data/association-flag-map.yaml
```diff
@@ -79,18 +79,15 @@ columns:
     - name: slot_Shape_flag
       bit: 23
       doc: General Failure Flag
-    - name: slot_Shape_flag_unweightedBad
+    - name: slot_Shape_flag_no_pixels
       bit: 24
-      doc: Both weighted and unweighted moments were invalid
-    - name: slot_Shape_flag_unweighted
+      doc: no pixels to measure
```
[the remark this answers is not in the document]
parejkoj: My thought is that if people are using the bits explicitly, it's a "buyer beware" situation. We have the flag unpacker to manage this by bitfield name, so what gets stored in a bit shouldn't matter.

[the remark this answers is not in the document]
kfindeisen: It would still matter when reading old files with new code. I assume it won't come up?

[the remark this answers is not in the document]
parejkoj: Do we write this output to files, or only to APDB?

Oh... I see your point though: changing the values here mean that old APDB instances cannot be interpreted. I think that's completely fine (double check with Eric though): a given APDB instance is tied to the Pipelines release it was generated with. This is how we'll handle schema changes as well.

Schema changes are more obvious to a user than changes in the meanings of bits, though. Do we have any way to version this bit map?

data/association-flag-map.yaml
kherner: We're not; I just copied all of the bits verbatim from the HSM shape code. See lines 45-49 of https://github.com/lsst/meas_extensions_shapeHSM/blob/main/src/HsmShape.cc

data/association-flag-map.yaml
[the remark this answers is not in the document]
kherner: Well we could do it by ap_association version, I suppose, though that's not perfect (albeit better than nothing).

data/association-flag-map.yaml
kfindeisen: Then I suggest removing this one -- even if it gets produced in the DiaSource catalog, there's no reason to copy it here.

kfindeisen (approved):

data/association-flag-map.yaml
kherner: I suppose we could only include mappings for the bits we're sure make sense, though it seemed easier to me at the time to just copy the entire thing for reasons of consistency.

kfindeisen: This was a few years back, but I think I remember Morrison saying that we didn't have many bits to spare. So it would be better to keep the flag map as compact as possible.

kherner:
Thinking about it some more, we could instead add the stuff from ap_association as additional bits instead of deleting the ones associated with SdssShape; that would allow for more backwards compatibility. The only one that overlaps is the general slot shape flag, so users would have to keep in mind that it might behave differently depending on what slot shape alogorithm was used in the processing. Of course we hope they have some idea of what they're actually processing! Doing that would take us to 33 total bits. Can I safely assume we have 64 to play with?

parejkoj:
The DiaSource flags field in sdm_schemas/apdb.yml is type `long`, so 64 bits sounds right.

parejkoj:
Using a different set of bits (at least for now) isn't a bad idea. Maybe with a note saying "bits X-Y were used for the SDSS source shape algorithm".

kherner:
I'll add something like, "Only valid when SdssShape algorithm is used" to the doc line for the "older" bits. I'll keep everything in the same position as they were, and then add the "new" ones as 28-32 (0-indexed).

kfindeisen:
How does the user know when a given algorithm was used?

kfindeisen:
Also, do you plan to merge today? We have a CI run with DM-37302 working for today, and I'm not sure whether to merge DM-37667 for tomorrow or wait until the day after.

kherner:
Running Jenkins one more time now just to be safe.

kherner:
Actually I got an error when doing a short pipetask run:

`Traceback (most recent call last):
  File "/sdf/data/rubin/user/kherner/DM-37622/ap_association/python/lsst/ap/association/transformDiaSourceCatalog.py", line 150, in _create_bit_pack_mappings
    self.inputSchema.find(bit['name'])
KeyError: "Field with name 'ext_shapeHSM_HsmSourceMoments_flag_unweightedBad' not found"
`
The _flag_unweightedBad is one of the SdssShape flags, so it looks like there isn't mechanism for doing things as we had hoped. It must expect they all exist for whatever is set as the slot shape flag.

kherner:
Jenkins build succeeded; merging shortly.

kherner:
> How does the user know when a given algorithm was used?

Oh, I just meant that it would be the defaults unless they change it in detectAndMeasure, and if they do change that, presumably they know what they're doing enough to make the corresponding changes in ap_association.

PR #168, opened 2023-04-18
DM-31924: update to new system for catalog ID generation

TallJimbo:


python/lsst/ap/association/diaForcedSource.py
```diff
@@ -123,9 +131,18 @@ def run(self,
 
         afw_dia_objects = self._convert_from_pandas(dia_objects)
 
-        idFactoryDiff = afwTable.IdFactory.makeSource(
-            diffim.info.id,
-            afwTable.IdFactory.computeReservedFromMaxBits(int(expIdBits)))
+        if expIdBits is None:
```
parejkoj, line 134:
```suggestion
        if expIdBits is not None:
```

python/lsst/ap/association/diaForcedSource.py
```diff
@@ -108,11 +110,17 @@ def run(self,
             Used to assure that the pipeline includes all diaObjects that were
             updated in case one falls on the edge of the CCD.
         expIdBits : `int`
-            Bit length of the exposure id.
+            Bit length of the exposure id.  Deprecated in favor of
```
parejkoj, line 113: Is there a ticket number for that removal that you could mention in this docstring and above the `if` below?

python/lsst/ap/association/diaPipe.py
```diff
@@ -298,9 +299,10 @@ def __init__(self, initInputs=None, **kwargs):
 
     def runQuantum(self, butlerQC, inputRefs, outputRefs):
         inputs = butlerQC.get(inputRefs)
-        expId, expBits = butlerQC.quantum.dataId.pack("visit_detector",
-                                                      returnMaxBits=True)
-        inputs["ccdExposureIdBits"] = expBits
+        inputs["idGenerator"] = self.config.idGenerator.apply(butlerQC.quantum.dataId)
+        # Need to set ccdExposureIdBits (now deprecated) to None and pass it,
+        # since there are non-optional positional arguments after it.
+        inputs["ccdExposureIdBits"] = None
```
parejkoj, line 305: Since we have to change the API here anyway, should we create the `idGenerator` inside of `run` instead, so nothing has to be passed in? I was always a bit skeptical of this modification of `inputs`.

python/lsst/ap/association/diaPipe.py
```diff
@@ -447,9 +457,12 @@ def run(self,
         diaForcedSources = self.diaForcedSource.run(
             diaCalResult.diaObjectCat,
             diaCalResult.updatedDiaObjects.loc[:, "diaObjectId"].to_numpy(),
+            # Passing ccdExposureIdBits here will take care of warning about
+            # its deprecation.
```
parejkoj, line 461: It took me a bit to sort out what this comment meant. Maybe rephrase to something like "Deprecation warning about ccdExposureIdBits is handled inside diaForcesource.run()."?

python/lsst/ap/association/transformDiaSourceCatalog.py
```diff
@@ -187,7 +189,7 @@ def run(self,
             Filter band of the science image.
         ccdVisitId : `int`
             Identifier for this detector+visit.
-        funcs : `lsst.pipe.tasks.functors.Functors`
+        funcs : `lsst.pipe.tasks.functors.Functors`, optional
             Functors to apply to the catalog's columns.
```
parejkoj, line 193: Missing the new `idGenerator` docstring.

parejkoj (commented):
I've wondered about our modification of `inputs` in `runQuantum` in all these Tasks, instead of generating the idFactory inside `run` itself. Maybe that will make more sense when I review the pipe_tasks ticket, but I wonder whether we could clean that up if we're needing to modify the API here anyway.

python/lsst/ap/association/diaPipe.py
TallJimbo: (see related replies in pipe_tasks and in the main PR thread)

python/lsst/ap/association/transformDiaSourceCatalog.py
TallJimbo: I think I must have ultimately intended to remove the `idGenerator` argument instead, since it isn't actually used in `run` and I didn't mark `ccdVisitId` as deprecated.  I think that's because it uses just the `ccdVisitId` (instead of e.g. also making an `lsst.afw.table.IdFactory`) and I was going for minimal changes.  If you'd prefer for me to deprecate `ccdVisitId` and pass an `idGenerator` instance for better consistency with other tasks, I can do that instead.

TallJimbo (commented):
> I wonder whether we could clean that up if we're needing to modify the API here anyway.

I don't see one or the other as clearly better here, so I'm inclined to just leave things as they are.  Passing the data ID is nice in that it moves more logic inside `run`, but passing the `IdGenerator` lets it do some extracting over the "is there a data ID" question, avoiding what may be a lot of `if data_id is None` checks in `run`.

python/lsst/ap/association/transformDiaSourceCatalog.py
parejkoj: Good question! I think since we don't need the packer itself and this operates per-detector/visit, we can keep it as-is. Also, all that's used for is a log message and setting the value in the table.

python/lsst/ap/association/transformDiaSourceCatalog.py
```diff
@@ -187,7 +188,7 @@ def run(self,
             Filter band of the science image.
         ccdVisitId : `int`
             Identifier for this detector+visit.
-        funcs : `lsst.pipe.tasks.functors.Functors`
+        funcs : `lsst.pipe.tasks.functors.Functors`, optional
```
parejkoj, line 191: I think we can just remove this `funcs` entirely: it's entirely unused: the code uses `self.funcs`, set in `__init__`.

PR #169, opened 2023-04-18
DM-38753: Use InMemoryDatasetHandle for test

timj:
With some other minor cleanups.

### DIFF
--- a/python/lsst/ap/association/skyBotEphemerisQuery.py
+++ b/python/lsst/ap/association/skyBotEphemerisQuery.py
@@ -167,8 +167,7 @@ class SkyBotEphemerisQueryTask(PipelineTask):
         """
         # Grab the visitInfo from the raw to get the information needed on the
         # full visit.
-        visitInfo = visitInfos[0].get(
-            datasetType=self.config.connections.visitInfos)
+        visitInfo = visitInfos[0].get()
 
         # Midpoint time of the exposure in JD
         expMidPointEPOCH = visitInfo.date.get(system=DateTime.JD, scale=DateTime.UTC)
--- a/tests/test_skyBotEphemerisQuery.py
+++ b/tests/test_skyBotEphemerisQuery.py
@@ -34,18 +34,6 @@ from lsst.utils import getPackageDir
 import lsst.utils.tests
 
 
-class MockDeferredDatasetHandle:
-    """A container that allows passing objects to methods that expect a
-    DeferredDatasetHandle.
-    """
-
-    def __init__(self, object):
-        self._internal = object
-
-    def get(self, *args, **kwargs):
-        return self._internal
-
-
 class TestSkyBotEphemerisQuery(unittest.TestCase):
 
     def setUp(self):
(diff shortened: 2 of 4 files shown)


python/lsst/ap/association/skyBotEphemerisQuery.py
```diff
@@ -167,8 +167,7 @@ def run(self, visitInfos, visit):
         """
         # Grab the visitInfo from the raw to get the information needed on the
         # full visit.
-        visitInfo = visitInfos[0].get(
-            datasetType=self.config.connections.visitInfos)
+        visitInfo = visitInfos[0].get()
```
timj, line 170: I'm a bit confused by this `run` method since it takes `list[VisitInfo]` but then only uses the first one. If it's only using the first one why is it deferring the get?

timj: Ah. Okay. So the graph builder asks for visits and so this can be given exposures from a multi-snap visit and since the boresight is the same for both snaps it knows it only needs to use the first one.

kfindeisen: Deferral was part of the initial commit on d8a3fe4a3635e046502243477b705f79551932d6. You'll have to ask the authors where that came from.

tests/test_skyBotEphemerisQuery.py
```diff
@@ -61,8 +62,7 @@ def requestReplace(input1, input2):
             """Junk wrapper for replacing the external internal call with an
             internel data load.
             """
-            with open(os.path.join(getPackageDir("ap_association"),
-                                   "tests",
+            with open(os.path.join(TESTDIR,
```
kfindeisen, line 65: Is there any particular reason to prefer the `__file__` approach other than convention?

python/lsst/ap/association/skyBotEphemerisQuery.py
timj: @eggls6 do you have a comment on the deferred loading? It doesn't matter for this PR but I am intrigued since it complicates the `run()` method to make it take a deferred handle and not a `VisitInfo`.

kfindeisen (approved):
Looks good!

tests/test_skyBotEphemerisQuery.py
timj: There are a couple of reasons why `__file__` is better.

* If the test data is in a subdir of the `tests` directory then you almost certainly want the data in that subdirectory to be the data that is used rather than whatever is pointed to by the setup package.
* `getPackageDir` relies on the environment variable being set (presumably by EUPS) and this is an unnecessary requirement when it comes to finding ones own test files in a subdirectory.

`getPackageDir` is used to find other packages and finding your own files relative to your own test directory shouldn't require EUPS. In an ideal world we would use `importlib.resources` to find data files and configs from other python packages but we haven't tried to tackle that yet.

I really thought we mentioned this in the dev guide but it seems there is no mention of how to find your own test files anywhere in the testing section.

kfindeisen: Sure, but you need the package to be set up anyway to be able to run the tests, so it seems to me like it amounts to the same thing. 🤷

timj: You could run the tests in a local directory but have the installed package be the one that is set up. Depends on what you are trying to do. Using `__file__` at least makes the tests be self-consistent without them causing confusion and finding the data from some other location.

PR #170, opened 2023-05-09
DM-37196: Update fields to reflect new APDB schema

parejkoj:


erinleighh (approved):
Everything looks good. Follows guidelines, and as far as I can tell, you found all the spots that needed to be updated. Not an easy feat!

PR #173, opened 2023-06-29
DM-38850: Make trailedAssociatorTask

bsmartradio:
Make trailedAssociatorTask which filters out trails whose lengths are above 0.416 arcseconds/second in length. The trailed sources are currently dropped once they are filtered out. More complexity for filtering will be added at a later date.

### DIFF
--- a/python/lsst/ap/association/association.py
+++ b/python/lsst/ap/association/association.py
@@ -101,16 +101,16 @@ class AssociationTask(pipeBase.Task):
         result : `lsst.pipe.base.Struct`
             Results struct with components.
 
-            - ``"matchedDiaSources"`` : DiaSources that were matched. Matched
+            - ``matchedDiaSources`` : DiaSources that were matched. Matched
               Sources have their diaObjectId updated and set to the id of the
               diaObject they were matched to. (`pandas.DataFrame`)
-            - ``"unAssocDiaSources"`` : DiaSources that were not matched.
+            - ``unAssocDiaSources`` : DiaSources that were not matched.
               Unassociated sources have their diaObject set to 0 as they
               were not associated with any existing DiaObjects.
               (`pandas.DataFrame`)
-            - ``"nUpdatedDiaObjects"`` : Number of DiaObjects that were
+            - ``nUpdatedDiaObjects`` : Number of DiaObjects that were
               matched to new DiaSources. (`int`)
-            - ``"nUnassociatedDiaObjects"`` : Number of DiaObjects that were
+            - ``nUnassociatedDiaObjects`` : Number of DiaObjects that were
               not matched a new DiaSource. (`int`)
         """
         diaSources = self.check_dia_source_radec(diaSources)
@@ -125,9 +125,8 @@ class AssociationTask(pipeBase.Task):
             diaTrailedResult = self.trailedSourceFilter.run(diaSources, exposure_time)
             matchResult = self.associate_sources(diaObjects, diaTrailedResult.diaSources)
 
-            self.log.warning("%i DIASources exceed maxTrailLength, dropping "
-                             "from source catalog."
-                             % len(diaTrailedResult.trailedDiaSources))
+            self.log.info("%i DIASources exceed max_trail_length, dropping "
+                          "from source catalog." % len(diaTrailedResult.trailedDiaSources))
 
         else:
             matchResult = self.associate_sources(diaObjects, diaSources)
@@ -189,11 +188,11 @@ class AssociationTask(pipeBase.Task):
         result : `lsst.pipe.base.Struct`
             Results struct with components.
 
-            - ``"diaSources"`` : Full set of diaSources both matched and not.
+            - ``diaSources`` : Full set of diaSources both matched and not.
               (`pandas.DataFrame`)
-            - ``"nUpdatedDiaObjects"`` : Number of DiaObjects that were
+            - ``nUpdatedDiaObjects`` : Number of DiaObjects that were
               associated. (`int`)
-            - ``"nUnassociatedDiaObjects"`` : Number of DiaObjects that were
+            - ``nUnassociatedDiaObjects`` : Number of DiaObjects that were
               not matched a new DiaSource. (`int`)
         """
         scores = self.score(
@@ -228,11 +227,11 @@ class AssociationTask(pipeBase.Task):
         result : `lsst.pipe.base.Struct`
             Results struct with components:
 
-            - ``"scores"``: array of floats of match quality updated DIAObjects
+            - ``scores``: array of floats of match quality updated DIAObjects
                 (array-like of `float`).
-            - ``"obj_idxs"``: indexes of the matched DIAObjects in the catalog.
+            - ``obj_idxs``: indexes of the matched DIAObjects in the catalog.
                 (array-like of `int`)
-            - ``"obj_ids"``: array of floats of match quality updated DIAObjects
+            - ``obj_ids``: array of floats of match quality updated DIAObjects
                 (array-like of `int`).
 
             Default values for these arrays are
--- a/python/lsst/ap/association/association.py
+++ b/python/lsst/ap/association/association.py
@@ -32,6 +32,7 @@ import lsst.geom as geom
 import lsst.pex.config as pexConfig
 import lsst.pipe.base as pipeBase
 from lsst.utils.timer import timeMethod
+from .trailedSourceFilter import TrailedSourceFilterTask
 
 # Enforce an error for unsafe column/array value setting in pandas.
 pd.options.mode.chained_assignment = 'raise'
@@ -40,11 +41,25 @@ pd.options.mode.chained_assignment = 'raise'
 class AssociationConfig(pexConfig.Config):
     """Config class for AssociationTask.
     """
+
     maxDistArcSeconds = pexConfig.Field(
         dtype=float,
-        doc='Maximum distance in arcseconds to test for a DIASource to be a '
-        'match to a DIAObject.',
+        doc="Maximum distance in arcseconds to test for a DIASource to be a "
+        "match to a DIAObject.",
         default=1.0,
+    )
+
+    trailedSourceFilter = pexConfig.ConfigurableField(
+        target=TrailedSourceFilterTask,
+        doc="Subtask to remove long trailed sources based on catalog source "
+            "morphological measurements.",
+    )
+
+    doTrailedSourceFilter = pexConfig.Field(
+        doc="Run traildeSourceFilter to remove long trailed sources from "
+            "output catalog.",
+        dtype=bool,
+        default=True,
     )
 
 
@@ -60,10 +75,16 @@ class AssociationTask(pipeBase.Task):
     ConfigClass = AssociationConfig
     _DefaultName = "association"
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if self.config.doTrailedSourceFilter:
+            self.makeSubtask("trailedSourceFilter")
+
     @timeMethod
     def run(self,
             diaSources,
-            diaObjects):
+            diaObjects,
+            exposure_time=None):
         """Associate the new DiaSources with existing DiaObjects.
 
         Parameters
@@ -72,6 +93,8 @@ class AssociationTask(pipeBase.Task):
             New DIASources to be associated with existing DIAObjects.
         diaObjects : `pandas.DataFrame`
             Existing diaObjects from the Apdb.
+        exposure_time : `float`, optional
+            Exposure time from difference image.
 
         Returns
         -------
@@ -98,7 +121,16 @@ class AssociationTask(pipeBase.Task):
                 nUpdatedDiaObjects=0,
                 nUnassociatedDiaObjects=0)
 
-        matchResult = self.associate_sources(diaObjects, diaSources)
+        if self.config.doTrailedSourceFilter:
+            diaTrailedResult = self.trailedSourceFilter.run(diaSources, exposure_time)
+            matchResult = self.associate_sources(diaObjects, diaTrailedResult.diaSources)
+
+            self.log.warning("%i DIASources exceed maxTrailLength, dropping "
+                             "from source catalog."
+                             % len(diaTrailedResult.trailedDiaSources))
+
+        else:
+            matchResult = self.associate_sources(diaObjects, diaSources)
 
         mask = matchResult.diaSources["diaObjectId"] != 0
 
--- a/python/lsst/ap/association/diaPipe.py
+++ b/python/lsst/ap/association/diaPipe.py
@@ -28,6 +28,10 @@ images at the updated locations of DiaObjects.
 Currently loads directly from the Apdb rather than pre-loading.
 """
 
+__all__ = ("DiaPipelineConfig",
+           "DiaPipelineTask",
+           "DiaPipelineConnections")
+
 import pandas as pd
 
 import lsst.dax.apdb as daxApdb
@@ -43,10 +47,6 @@ from lsst.ap.association import (
     LoadDiaCatalogsTask,
     PackageAlertsTask)
 from lsst.ap.association.ssoAssociation import SolarSystemAssociationTask
-
-__all__ = ("DiaPipelineConfig",
-           "DiaPipelineTask",
-           "DiaPipelineConnections")
 
 
 class DiaPipelineConnections(
@@ -367,8 +367,8 @@ class DiaPipelineTask(pipeBase.PipelineTask):
         loaderResult = self.diaCatalogLoader.run(diffIm, self.apdb)
 
         # Associate new DiaSources with existing DiaObjects.
-        assocResults = self.associator.run(diaSourceTable,
-                                           loaderResult.diaObjects)
+        assocResults = self.associator.run(diaSourceTable, loaderResult.diaObjects,
+                                           exposure_time=diffIm.getInfo().getVisitInfo().getExposureTime())
         if self.config.doSolarSystemAssociation:
             ssoAssocResult = self.solarSystemAssociator.run(
                 assocResults.unAssocDiaSources,
--- a/python/lsst/ap/association/diaPipe.py
+++ b/python/lsst/ap/association/diaPipe.py
@@ -368,7 +368,7 @@ class DiaPipelineTask(pipeBase.PipelineTask):
 
         # Associate new DiaSources with existing DiaObjects.
         assocResults = self.associator.run(diaSourceTable, loaderResult.diaObjects,
-                                           exposure_time=diffIm.getInfo().getVisitInfo().getExposureTime())
+                                           exposure_time=diffIm.visitInfo.exposureTime)
         if self.config.doSolarSystemAssociation:
             ssoAssocResult = self.solarSystemAssociator.run(
                 assocResults.unAssocDiaSources,
--- a/python/lsst/ap/association/trailedSourceFilter.py
+++ b/python/lsst/ap/association/trailedSourceFilter.py
@@ -30,12 +30,12 @@ class TrailedSourceFilterConfig(pexConfig.Config):
     """Config class for TrailedSourceFilterTask.
     """
 
-    maxTrailLength = pexConfig.Field(
+    max_trail_length = pexConfig.Field(
         dtype=float,
         doc="Length of long trailed sources to remove from the input catalog, "
             "in arcseconds per second. Default comes from DMTN-199, which "
             "requires removal of sources with trails longer than 10 "
-            "degrees/day, which is 36000/3600/24arcsec/second, or roughly"
+            "degrees/day, which is 36000/3600/24 arcsec/second, or roughly"
             "0.416 arcseconds per second.",
         default=36000/3600.0/24.0,
     )
@@ -46,9 +46,9 @@ class TrailedSourceFilterTask(pipeBase.Task):
     guidelines.
 
     This task checks the length of trailLength in the DIASource catalog using
-    a given arcsecond/second rate from maxTrailLength and the exposure time.
+    a given arcsecond/second rate from max_trail_length and the exposure time.
     The two values are used to calculate the maximum allowed trail length and
-    filters out any trail longer than the maximum. The maxTrailLength is
+    filters out any trail longer than the maximum. The max_trail_length is
     outlined in DMTN-199 and determines the default value.
     """
 
@@ -57,7 +57,7 @@ class TrailedSourceFilterTask(pipeBase.Task):
 
     @timeMethod
     def run(self, dia_sources, exposure_time):
-        """Remove trailed sources longer than ``config.maxTrailLength`` from
+        """Remove trailed sources longer than ``config.max_trail_length`` from
         the input catalog.
 
         Parameters
@@ -76,10 +76,9 @@ class TrailedSourceFilterTask(pipeBase.Task):
              trailed sources. (`pandas.DataFrame`)
 
             - ``trailed_dia_sources`` : DIASources that have trails which
-            exceed maxTrailLength/second*exposure_time.
+            exceed max_trail_length/second*exposure_time.
             (`pandas.DataFrame`)
         """
-
         trail_mask = self._check_dia_source_trail(dia_sources, exposure_time)
 
         return pipeBase.Struct(
@@ -89,8 +88,8 @@ class TrailedSourceFilterTask(pipeBase.Task):
     def _check_dia_source_trail(self, dia_sources, exposure_time):
         """Find DiaSources that have long trails.
 
-        Creates a mask for sources with lengths greater than 0.416
-        arcseconds/second multiplied by the exposure time.
+        Return a mask of sources with lengths greater than
+        ``config.max_trail_length``  multiplied by the exposure time.
 
         Parameters
         ----------
@@ -105,8 +104,7 @@ class TrailedSourceFilterTask(pipeBase.Task):
             Boolean mask for DIASources which are greater than the
             cutoff length.
         """
-
         trail_mask = (dia_sources.loc[:, "trailLength"].values[:]
-                      >= (self.config.maxTrailLength*exposure_time))
+                      >= (self.config.max_trail_length*exposure_time))
 
         return trail_mask
--- a/python/lsst/ap/association/trailedSourceFilter.py
+++ b/python/lsst/ap/association/trailedSourceFilter.py
@@ -1,0 +1,112 @@
+# This file is part of ap_association.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+__all__ = ("TrailedSourceFilterTask", "TrailedSourceFilterConfig")
+
+import lsst.pex.config as pexConfig
+import lsst.pipe.base as pipeBase
+from lsst.utils.timer import timeMethod
+
+
+class TrailedSourceFilterConfig(pexConfig.Config):
+    """Config class for TrailedSourceFilterTask.
+    """
+
+    maxTrailLength = pexConfig.Field(
+        dtype=float,
+        doc="Length of long trailed sources to remove from the input catalog, "
+            "in arcseconds per second. Default comes from DMTN-199, which "
+            "requires removal of sources with trails longer than 10 "
+            "degrees/day, which is 36000/3600/24arcsec/second, or roughly"
+            "0.416 arcseconds per second.",
+        default=36000/3600.0/24.0,
+    )
+
+
+class TrailedSourceFilterTask(pipeBase.Task):
+    """Find trailed sources in DIASources and filter them as per DMTN-199
+    guidelines.
+
+    This task checks the length of trailLength in the DIASource catalog using
+    a given arcsecond/second rate from maxTrailLength and the exposure time.
+    The two values are used to calculate the maximum allowed trail length and
+    filters out any trail longer than the maximum. The maxTrailLength is
+    outlined in DMTN-199 and determines the default value.
+    """
+
+    ConfigClass = TrailedSourceFilterConfig
+    _DefaultName = "trailedSourceFilter"
+
+    @timeMethod
+    def run(self, dia_sources, exposure_time):
+        """Remove trailed sources longer than ``config.maxTrailLength`` from
+        the input catalog.
+
+        Parameters
+        ----------
+        dia_sources : `pandas.DataFrame`
+            New DIASources to be checked for trailed sources.
+        exposure_time : `float`
+            Exposure time from difference image.
+
+        Returns
+        -------
+        result : `lsst.pipe.base.Struct`
+            Results struct with components.
+
+            - ``dia_sources`` : DIASource table that is free from unwanted
+             trailed sources. (`pandas.DataFrame`)
+
+            - ``trailed_dia_sources`` : DIASources that have trails which
+            exceed maxTrailLength/second*exposure_time.
+            (`pandas.DataFrame`)
+        """
+
+        trail_mask = self._check_dia_source_trail(dia_sources, exposure_time)
+
+        return pipeBase.Struct(
+            diaSources=dia_sources[~trail_mask].reset_index(drop=True),
+            trailedDiaSources=dia_sources[trail_mask].reset_index(drop=True))
+
+    def _check_dia_source_trail(self, dia_sources, exposure_time):
+        """Find DiaSources that have long trails.
+
+        Creates a mask for sources with lengths greater than 0.416
+        arcseconds/second multiplied by the exposure time.
+
+        Parameters
+        ----------
+        dia_sources : `pandas.DataFrame`
+            Input DIASources to check for trail lengths.
+        exposure_time : `float`
+            Exposure time from difference image.
+
+        Returns
+        -------
+        trail_mask : `pandas.DataFrame`
+            Boolean mask for DIASources which are greater than the
+            cutoff length.
+        """
+
+        trail_mask = (dia_sources.loc[:, "trailLength"].values[:]
+                      >= (self.config.maxTrailLength*exposure_time))
+
+        return trail_mask
--- a/tests/test_association_task.py
+++ b/tests/test_association_task.py
@@ -22,7 +22,6 @@
 import numpy as np
 import pandas as pd
 import unittest
-
 import lsst.geom as geom
 import lsst.utils.tests
 
@@ -46,20 +45,23 @@ class TestAssociationTask(unittest.TestCase):
         self.diaSources = pd.DataFrame(data=[
             {"ra": 0.04*idx + scatter*rng.uniform(-1, 1),
              "dec": 0.04*idx + scatter*rng.uniform(-1, 1),
-             "diaSourceId": idx + 1 + self.nObjects, "diaObjectId": 0}
+             "diaSourceId": idx + 1 + self.nObjects, "diaObjectId": 0, "trailLength": 5.5*idx}
             for idx in range(self.nSources)])
         self.diaSourceZeroScatter = pd.DataFrame(data=[
             {"ra": 0.04*idx,
              "dec": 0.04*idx,
-             "diaSourceId": idx + 1 + self.nObjects, "diaObjectId": 0}
+             "diaSourceId": idx + 1 + self.nObjects, "diaObjectId": 0, "trailLength": 5.5*idx}
             for idx in range(self.nSources)])
+        self.exposure_time = 30.0
 
     def test_run(self):
         """Test the full task by associating a set of diaSources to
         existing diaObjects.
         """
-        assocTask = AssociationTask()
-        results = assocTask.run(self.diaSources, self.diaObjects)
+        config = AssociationTask.ConfigClass()
+        config.doTrailedSourceFilter = False
+        assocTask = AssociationTask(config=config)
+        results = assocTask.run(self.diaSources, self.diaObjects, exposure_time=self.exposure_time)
 
         self.assertEqual(results.nUpdatedDiaObjects, len(self.diaObjects) - 1)
         self.assertEqual(results.nUnassociatedDiaObjects, 1)
@@ -75,13 +77,31 @@ class TestAssociationTask(unittest.TestCase):
                 [0]):
             self.assertEqual(test_obj_id, expected_obj_id)
 
+    def test_run_trailed_sources(self):
+        """Test the full task by associating a set of diaSources to
+        existing diaObjects when trailed sources are filtered.
+
+        This should filter out two of the five sources based on trail length,
+        leaving one unassociated diaSource and two associated diaSources.
+        """
+        assocTask = AssociationTask()
+        results = assocTask.run(self.diaSources, self.diaObjects, exposure_time=self.exposure_time)
+
+        self.assertEqual(results.nUpdatedDiaObjects, len(self.diaObjects) - 3)
+        self.assertEqual(results.nUnassociatedDiaObjects, 3)
+        self.assertEqual(len(results.matchedDiaSources), len(self.diaObjects) - 3)
+        self.assertEqual(len(results.unAssocDiaSources), 1)
+        np.testing.assert_array_equal(results.matchedDiaSources["diaObjectId"].values, [1, 2])
+        np.testing.assert_array_equal(results.unAssocDiaSources["diaObjectId"].values, [0])
+
     def test_run_no_existing_objects(self):
         """Test the run method with a completely empty database.
         """
         assocTask = AssociationTask()
         results = assocTask.run(
             self.diaSources,
-            pd.DataFrame(columns=["ra", "dec", "diaObjectId"]))
+            pd.DataFrame(columns=["ra", "dec", "diaObjectId", "trailLength"]),
+            exposure_time=self.exposure_time)
         self.assertEqual(results.nUpdatedDiaObjects, 0)
         self.assertEqual(results.nUnassociatedDiaObjects, 0)
         self.assertEqual(len(results.matchedDiaSources), 0)
--- a/tests/test_association_task.py
+++ b/tests/test_association_task.py
@@ -68,14 +68,8 @@ class TestAssociationTask(unittest.TestCase):
         self.assertEqual(len(results.matchedDiaSources),
                          len(self.diaObjects) - 1)
         self.assertEqual(len(results.unAssocDiaSources), 1)
-        for test_obj_id, expected_obj_id in zip(
-                results.matchedDiaSources["diaObjectId"].to_numpy(),
-                [1, 2, 3, 4]):
-            self.assertEqual(test_obj_id, expected_obj_id)
-        for test_obj_id, expected_obj_id in zip(
-                results.unAssocDiaSources["diaObjectId"].to_numpy(),
-                [0]):
-            self.assertEqual(test_obj_id, expected_obj_id)
+        np.testing.assert_array_equal(results.matchedDiaSources["diaObjectId"].values, [1, 2, 3, 4])
+        np.testing.assert_array_equal(results.unAssocDiaSources["diaObjectId"].values, [0])
 
     def test_run_trailed_sources(self):
         """Test the full task by associating a set of diaSources to
@@ -119,6 +113,7 @@ class TestAssociationTask(unittest.TestCase):
                 assoc_result.diaSources["diaObjectId"].to_numpy(),
                 [0, 1, 2, 3, 4]):
             self.assertEqual(test_obj_id, expected_obj_id)
+        np.testing.assert_array_equal(assoc_result.diaSources["diaObjectId"].values, [0, 1, 2, 3, 4])
 
     def test_score_and_match(self):
         """Test association between a set of sources and an existing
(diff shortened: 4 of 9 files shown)


parejkoj:
The pull request should have the ticket number in it: https://developer.lsst.io/work/flow.html#make-a-pull-request

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class TrailedSourceFilterConfig(pexConfig.Config):
+    """Config class for TrailedSourceFilterTask.
+    """
+    maxTrailLength = pexConfig.Field(
+        dtype=float,
+        doc='Maximum trail length permitted is less than 10 degrees/day. This is a rate '
+            'of 0.416 arcseconds per second.As trail length is measured in '
+            'arcseconds, it is dependant on the length of the exposure.',
```
parejkoj, line 45: Please make the docstring describe what `maxTrailLength` does, not what its default value means. You can also include a note about that default value, but only after describing what it does.

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,110 @@
...
+        default=0.416,  # HSC Default. Should Decam and LSST defaults be passed? Does it change possibly?
+    )
+
+
+class TrailedSourceFilterTask(pipeBase.Task):
+    """Find trailed sources in DIAObjects.
+
```
parejkoj, line 52: Please add an extended summary that discusses what this task does. It also doesn't operate on DiaObjects, but rather on Diasources and an exposure.

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,110 @@
...
+    """
+
+    ConfigClass = TrailedSourceFilterConfig
+    _DefaultName = "trailedAssociation"
```
parejkoj, line 56: This should be `trailedSourceFilter`

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,110 @@
...
+
+    @timeMethod
+    def run(self,
+            dia_sources, exposure):
```
parejkoj, line 60: formatting?

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,110 @@
...
+        """Find trailed sources which have not been filtered out and will
+        not be included in the diaSource catalog.
```
parejkoj, line 62: "which have not been filtered out"? What do you mean by that?

parejkoj, line 62: "and will not be included in the diaSource catalog" -> "to not include in the diaSource catalog"?

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,110 @@
...
+
+        Parameters
+        ----------
+        dia_sources : `pandas.DataFrame`
+            New DIASources to be checked for trailed sources.
+        exposure : `pandas.DataFrame`
```
parejkoj, line 68: Since this is the difference image, maybe call it `difference` instead of `exposure`, to make it clear it's not just any exposure.

Also, it's not a DataFrame.

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,110 @@
...
+            Calibrated exposure differenced with a template image during
+            image differencing.
+
+        Returns
+        -------
+        result : `lsst.pipe.base.Struct`
+            Results struct with components.
+
+            - ``"dia_sources"`` : DiaSource table that is free from unwanted
```
parejkoj, line 77: No inner double quotes when documenting structs.

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,110 @@
...
+             trailed sources (`pandas.DataFrame`)
+
+            - ``"trailed_dia_sources"`` : DiaSources that have trailed more than
+             0.416 arcseconds/second*exposure_time(`pandas.DataFrame`)
```
parejkoj, line 81: Do not include the default values here. I think you want this, but I'm not sure:

```
DiaSources that are considered to have trails longer than ``config.maxTrailLength``.
```

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,110 @@
...
+            image differencing.
+
+        Returns
+        -------
+        result : `lsst.pipe.base.Struct`
+            Results struct with components.
+
+            - ``"dia_sources"`` : DiaSource table that is free from unwanted
+             trailed sources (`pandas.DataFrame`)
+
```
parejkoj, line 79: no newline here

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,110 @@
...
+            - ``"trailed_dia_sources"`` : DiaSources that have trailed more than
+             0.416 arcseconds/second*exposure_time(`pandas.DataFrame`)
+        """
+        trail_mask = self.check_dia_source_trail(dia_sources, exposure)
+
+        return pipeBase.Struct(
+            diaSources=dia_sources[~trail_mask].reset_index(drop=True),
+            trailedDiaSources=dia_sources[trail_mask].reset_index(drop=True))
+
+    def check_dia_source_trail(self, dia_sources, exposure):
```
parejkoj, line 89: This should be private (prefix with `_`): we didn't do that with older tasks, and have come to regret it. The only public entry point for tasks should be `run`.

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,110 @@
...
+        """Check that all DiaSources have trails.
```
parejkoj, line 90:
```suggestion
        """Find DiaSources that have trails.
```

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,110 @@
...
+
+        Creates a mask for sources with lengths greater than 0.416 arcseconds/second*exposure time.
```
parejkoj, line 92: Again, do not mention the default values here.

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,110 @@
...
+
+        Parameters
+        ----------
+        dia_sources : `pandas.DataFrame`
+            Input DiaSources to check for trail lengths.
+        exposure : `pandas.DataFrame`
```
parejkoj, line 98: Not a DataFrame, and probably call it `difference`.

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,110 @@
...
+            Calibrated exposure differenced with a template image during
+            image differencing.
+
+        Returns
+        -------
+        trail_mask : `pandas.DataFrame`
+            Boolean mask for dia_sources which are greater than the cuttoff length.
+        """
+        exposure_time = exposure.getInfo().getVisitInfo().getExposureTime()
```
parejkoj, line 107: The only thing you use the exposure for here is the exposure time? In that case, you should just be passing in the time (to this method, and `run`), as you don't need the rest of the exposure at all. There's no pixel math happening here, so all we need is the time (and note in the docstrings that it's in seconds).

tests/test_association_task.py
```diff
@@ -46,20 +48,38 @@ def setUp(self):
...
+        visitInfo = afwImage.VisitInfo(
+            exposureId=exposureId,
+            exposureTime=exposureTime,
+            boresightRotAngle=boresightRotAngle,
+            weather=weather,
+        )
+        exposureInfo = afwImage.ExposureInfo()
+        exposureInfo.setVisitInfo(visitInfo)
+        maskedImage = afwImage.MaskedImageF(lsst.geom.Extent2I(64, 64))
+        self.exposure = afwImage.ExposureF(maskedImage, exposureInfo)
```
parejkoj, line 71: See my comments above: you don't need an exposure at all, just a time in seconds, so this is all unnecessary.

tests/test_association_task.py
```diff
@@ -46,20 +48,38 @@ def setUp(self):
...
 
     def test_run(self):
         """Test the full task by associating a set of diaSources to
         existing diaObjects.
-        """
-        assocTask = AssociationTask()
-        results = assocTask.run(self.diaSources, self.diaObjects)
+        # """
```
parejkoj, line 76: What happened here?

tests/test_association_task.py
```diff
@@ -75,13 +95,40 @@ def test_run(self):
                 [0]):
             self.assertEqual(test_obj_id, expected_obj_id)
 
+    def test_run_trailed_sources(self):
+        """Test the full task by associating a set of diaSources to
+        existing diaObjects when trailed sources are filtered
+        # """
```
parejkoj, line 101: ???

tests/test_association_task.py
```diff
@@ -75,13 +95,40 @@ def test_run(self):
                 [0]):
             self.assertEqual(test_obj_id, expected_obj_id)
 
+    def test_run_trailed_sources(self):
+        """Test the full task by associating a set of diaSources to
+        existing diaObjects when trailed sources are filtered
```
parejkoj, line 100: Period at end of sentence.

tests/test_association_task.py
```diff
@@ -75,13 +95,40 @@ def test_run(self):
...
+        # """
+
+        config = AssociationTask.ConfigClass()
+        config.doLongTrailFilter = True
```
parejkoj, line 104: Since it's the default, you don't need to set it here, or even create a config at all.

tests/test_association_task.py
```diff
@@ -100,6 +147,9 @@ def test_associate_sources(self):
                 [0, 1, 2, 3, 4]):
             self.assertEqual(test_obj_id, expected_obj_id)
 
+    def test_trailed_source(self):
+        """Test the removal of trailed sources"""
```
parejkoj, line 151: This does nothing.

tests/test_association_task.py
```diff
@@ -75,13 +95,40 @@ def test_run(self):
...
+        # """
+
+        config = AssociationTask.ConfigClass()
+        config.doLongTrailFilter = True
+        assocTask = AssociationTask(config=config)
+
+        results = assocTask.run(self.diaSources, self.diaObjects, self.exposure)
+
+        print(results.nUpdatedDiaObjects, "results nupdated")
+        print(len(self.diaObjects) - 3, " length of dia objects")
```
parejkoj, line 110: No print statements in task or test code.

tests/test_association_task.py
```diff
@@ -75,13 +95,40 @@ def test_run(self):
...
+        self.assertEqual(len(results.unAssocDiaSources), 1)
+        for test_obj_id, expected_obj_id in zip(
+                results.matchedDiaSources["diaObjectId"].to_numpy(),
+                [1, 2, 3, 4]):
+            self.assertEqual(test_obj_id, expected_obj_id)
+        for test_obj_id, expected_obj_id in zip(
+                results.unAssocDiaSources["diaObjectId"].to_numpy(),
+                [0]):
+            self.assertEqual(test_obj_id, expected_obj_id)
+
```
parejkoj, line 124: This test block looks very similar to the one in `test_run`: how many of these values are actually different? How many trailed sources do you expect to have in the test dataset, and does that match the changes here?

I think you'd be better off having a separate set of unittests of the new Task itself, in its own test file, so you can confirm that the new Task's `run` is returning the correct things in its Struct.

parejkoj (changes requested):
Have you run this? I think it would have failed a linter, at least. There were a lot of typos and leftover print/import statements that were clearly from debugging; please try to have those cleaned up before you send it for review.

There are no tests specific to the new Task itself: please add a file to test the new Task output.

Is this really worth having a separate Task, instead of just adding a config, branch, and method to `AssociationTask`? It's just ~3 lines of code; do we expect that we will be significantly expanding functionality compared to what you've implemented here, and are there any other places where we need this as a Task?

tests/test_association_task.py
bsmartradio: Yup, forgot to remove that after testing. Removed now.

python/lsst/ap/association/association.py
```diff
@@ -47,10 +48,22 @@ class AssociationConfig(pexConfig.Config):
         default=1.0,
     )
 
+    trailedSourceFilter = pexConfig.ConfigurableField(
+        target=TrailedSourceFilterTask,
+        doc="Docs here"
+
+    )
+
+    doLongTrailFilter = pexConfig.Field(
```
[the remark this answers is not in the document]
bsmartradio: Changed name to `doTrailedSourceFilter`

python/lsst/ap/association/association.py
```diff
@@ -47,10 +48,22 @@ class AssociationConfig(pexConfig.Config):
         default=1.0,
     )
 
+    trailedSourceFilter = pexConfig.ConfigurableField(
+        target=TrailedSourceFilterTask,
+        doc="Docs here"
```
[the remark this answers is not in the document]
bsmartradio: Moced the docstring for `doTrailedSourceFilter` filter to the `trailedSourceFilter` and changed the `doTrailedSourceFilter` docstring to one reflecting that it is setting the subtask flag to true or false.

python/lsst/ap/association/association.py
```diff
@@ -60,10 +73,16 @@ class AssociationTask(pipeBase.Task):
...
+        super().__init__(*args, **kwargs)
+        if self.config.doLongTrailFilter:
+            self.makeSubtask("trailedSourceFilter")
+
     @timeMethod
     def run(self,
             diaSources,
-            diaObjects):
+            diaObjects,
+            exposure):
```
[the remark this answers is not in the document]
bsmartradio: Changed `expsoure` to `exposure=exposure` and updated the docs to reflect that it is an optional parameter.

tests/test_association_task.py
bsmartradio: Removed extra test.

tests/test_association_task.py
bsmartradio: I've created a new unit test called `test_trailedSourceFilter.py`. I wasn't sure if I should leave testing it out completely from the association tests because filtering out does change the expected number of associated sources and objects. As of right now, we haven't run this on a full dataset so it's hard to say how many sources will be filtered out. I'm hoping Erin can do a test run of this before it's merged onto main so we can see the effects. since ap_verify doesn't have any sources that meet the filter criteria.

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,110 @@
...
+
+import lsst.pex.config as pexConfig
+import lsst.pipe.base as pipeBase
+from lsst.utils.timer import timeMethod
+
+# import numpy as np
+import pandas as pd
+
+# Enforce an error for unsafe column/array value setting in pandas.
+pd.options.mode.chained_assignment = 'raise'
```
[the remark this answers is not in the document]
bsmartradio: I followed what was being done in association.py. If it applies to all code run after, then since this runs after association.py has imported and run it would apply to the trailedSourceFilter as well. However, with what you said, should we reconsider it being set in association.py? What would be a better way to set it?

python/lsst/ap/association/trailedSourceFilter.py
bsmartradio: Added.

python/lsst/ap/association/diaPipe.py
```diff
@@ -28,6 +28,10 @@
 Currently loads directly from the Apdb rather than pre-loading.
 """
 
+__all__ = ("DiaPipelineConfig",
+           "DiaPipelineTask",
+           "DiaPipelineConnections")
```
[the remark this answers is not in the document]
bsmartradio: Looking at it, this was already there just below the imports, so I moved it above them.

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,110 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+"""A simple implementation of source association task for ap_verify.
+"""
+
+__all__ = ["TrailedSourceFilterTask", "TrailedSourceFilterConfig"]
```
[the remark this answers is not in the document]
bsmartradio: I had followed the example of association.py, and it looks like it uses a list as well, as does diaForcedSource, skyBotEphemeris, metrics, and ssoAssociation. Should I make a separate ticket to update all of these?

python/lsst/ap/association/trailedSourceFilter.py
bsmartradio: I followed the style in association.py. Should those doc strings be updated as well?

bsmartradio:
Here is the jenkins run using this branch.
https://rubin-ci.slac.stanford.edu/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/95/pipeline/

bsmartradio:
@parejkoj Are you happy with the changes? If so I can rebase onto main and merge the changes.

python/lsst/ap/association/association.py
```diff
@@ -72,6 +91,8 @@ def run(self,
             New DIASources to be associated with existing DIAObjects.
         diaObjects : `pandas.DataFrame`
             Existing diaObjects from the Apdb.
+        diffIm : `lsst.afw.image.ExposureF` optional
```
parejkoj, line 94:
```suggestion
        diffIm : `lsst.afw.image.ExposureF`, optional
```

python/lsst/ap/association/association.py
```diff
@@ -98,7 +119,13 @@ def run(self,
                 nUpdatedDiaObjects=0,
                 nUnassociatedDiaObjects=0)
 
-        matchResult = self.associate_sources(diaObjects, diaSources)
+        if self.config.doTrailedSourceFilter:
+            diaTrailedResult = self.trailedSourceFilter.run(diaSources, diffIm)
+
+            matchResult = self.associate_sources(diaObjects, diaTrailedResult.diaSources)
+
```
parejkoj, line 126: Why the extra newlines? There's no need to separate these.

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,106 @@
...
+
+class TrailedSourceFilterConfig(pexConfig.Config):
+    """Config class for TrailedSourceFilterTask.
+    """
+    maxTrailLength = pexConfig.Field(
+        dtype=float,
+        doc='Maximum trail length permitted is less than 10 degrees/day. '
+            'This is a rate of 0.416 arcseconds per second. As trail length'
+            ' is measured in arcseconds, it is dependant on the length of '
+            'the exposure.',
```
parejkoj, line 37: I'd reword this to lead with the units definition and meaning instead:

"Length of long trailed sources to remove from the input catalog, in arcseconds per second. Default comes from DMTN-199, which requires removal of sources with trails longer than 10 degrees/day, which is 0.416 arcsec/second."

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,106 @@
...
+        default=0.416,
+    )
+
+
+class TrailedSourceFilterTask(pipeBase.Task):
+    """Find trailed sources in DIASources. This task checks the length of
+    trailLength in the DIASource catalog using a given
+     arcsecond/second rate from maxTrailLength and the exposure time. The
+     two values are used to calculate the maximum allowed trail length and
+     filters out any trail longer than the maximum.
```
parejkoj, line 47: Inconsistent indentation.

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,106 @@
...
+
+    """
+
+    ConfigClass = TrailedSourceFilterConfig
+    _DefaultName = "trailedSourceFilter"
+
+    @timeMethod
+    def run(self, dia_sources, diffIm):
+        """Find trailed sources which do not meet requirements and
+        will not be included in the diaSource catalog.
```
parejkoj, line 57:
```suggestion
        """Remove trailed sources longer than ``config.maxTrailLength`` from the input catalog.
```

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,106 @@
...
+     two values are used to calculate the maximum allowed trail length and
+     filters out any trail longer than the maximum.
+
+    """
+
+    ConfigClass = TrailedSourceFilterConfig
+    _DefaultName = "trailedSourceFilter"
+
+    @timeMethod
+    def run(self, dia_sources, diffIm):
```
parejkoj, line 55: You're not consistent about snake vs. camelCase; per [RFC-623](https://jira.lsstcorp.org/browse/RFC-623) we can use snake case in new code, which this is. Please update to use snake_case throughout.

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,106 @@
...
+        diffIm : `lsst.afw.image.ExposureF`
+            Difference image on which the DIASources were detected.
+
+        Returns
+        -------
+        result : `lsst.pipe.base.Struct`
+            Results struct with components.
+
+            - ``dia_sources`` : DIASource table that is free from unwanted
+             trailed sources. (`pandas.DataFrame`)
```
parejkoj, line 72:
```suggestion
            - ``dia_sources`` : DIASource table that has had long trailed sources removed. (`pandas.DataFrame`)
```

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,106 @@
...
+
+            - ``trailed_dia_sources`` : DIASources that have trailed more
+            than 0.416 arcseconds/second*exposure_time. (`pandas.DataFrame`)
```
parejkoj, line 75:
```suggestion
            than maxTrailLength*exposure_time. (`pandas.DataFrame`)
```

We don't want to include numbers here, since it's configurable.

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,106 @@
...
+        """
+
```
parejkoj, line 77: No newlines after function docstrings. https://developer.lsst.io/python/numpydoc.html#docstrings-of-methods-and-functions-should-not-be-preceded-or-followed-by-a-blank-line

python/lsst/ap/association/association.py
```diff
@@ -47,6 +48,18 @@ class AssociationConfig(pexConfig.Config):
...
     )
 
+    trailedSourceFilter = pexConfig.ConfigurableField(
+        target=TrailedSourceFilterTask,
+        doc="Subtask to filter artifact candidates based on morphological "
+            "criteria, i.g. those that appear to be streaks.",
+    )
+
+    doTrailedSourceFilter = pexConfig.Field(
+        doc="Set flag for trailed source filter subtask to run.",
```
parejkoj, line 58:
```suggestion
        doc="Run trailedSourceFilter to remove long trailed sources from output catalog?.",
```

python/lsst/ap/association/association.py
```diff
@@ -47,6 +48,18 @@ class AssociationConfig(pexConfig.Config):
         default=1.0,
     )
 
+    trailedSourceFilter = pexConfig.ConfigurableField(
+        target=TrailedSourceFilterTask,
+        doc="Subtask to filter artifact candidates based on morphological "
+            "criteria, i.g. those that appear to be streaks.",
```
parejkoj, line 54: I don't think this is "artifact candidates"; the only thing that task does is remove long trails per DMTN-199.

```suggestion
        doc="Subtask to remove long trailed sources based on catalog source morphological measurements.",
```

python/lsst/ap/association/association.py
```diff
@@ -98,7 +120,18 @@ def run(self,
...
                 nUnassociatedDiaObjects=0)
 
-        matchResult = self.associate_sources(diaObjects, diaSources)
+        if self.config.doLongTrailFilter:
+            diaTrailedResult = self.trailedSourceFilter.run(diaSources, exposure)
+
+            if len(diaTrailedResult.trailedDiaSources) > 0:
+                print("Trailed sources cleaned.")
+            else:
+                print("No trailed sources to clean.")
```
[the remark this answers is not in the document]
parejkoj: It looks like you removed print statements, but didn't add a log.info: I think we absolutely want to have a log statement about how many sources were removed.

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,106 @@
...
+        doc='Maximum trail length permitted is less than 10 degrees/day. '
+            'This is a rate of 0.416 arcseconds per second. As trail length'
+            ' is measured in arcseconds, it is dependant on the length of '
+            'the exposure.',
+        default=0.416,
+    )
+
+
+class TrailedSourceFilterTask(pipeBase.Task):
+    """Find trailed sources in DIASources. This task checks the length of
```
parejkoj, line 43: I'd mention DMTN-199 in the opening sentence, since that is what is driving the implementation of this task.

python/lsst/ap/association/trailedSourceFilter.py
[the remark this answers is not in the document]
parejkoj: KT has gone the other way on my dev guide suggestion, so lets not touch anything else right now while we sort that out.

https://github.com/lsst-dm/dm_dev_guide/pull/632

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,106 @@
...
+class TrailedSourceFilterConfig(pexConfig.Config):
+    """Config class for TrailedSourceFilterTask.
+    """
+    maxTrailLength = pexConfig.Field(
+        dtype=float,
+        doc='Maximum trail length permitted is less than 10 degrees/day. '
+            'This is a rate of 0.416 arcseconds per second. As trail length'
+            ' is measured in arcseconds, it is dependant on the length of '
+            'the exposure.',
+        default=0.416,
```
parejkoj, line 38: I'd put this in explicitly as `10.0/3600/24`, since `0.416` is not the exact unit converted value.

python/lsst/ap/association/trailedSourceFilter.py
parejkoj: This is how they should be documented: https://developer.lsst.io/python/numpydoc.html#struct-types

Please file a ticket to go through all of `ap_assocation` and fix the `Struct` docstrings, if you know some are wrong.

python/lsst/ap/association/trailedSourceFilter.py
parejkoj: Why do you need the full exposure, and not just the exposure time here?

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,106 @@
...
+        diffIm : `pandas.DataFrame`
+            Difference image on which the DIASources were detected.
+
+        Returns
+        -------
+        trail_mask : `pandas.DataFrame`
+            Boolean mask for DIASources which are greater than the
+            cutoff length.
+        """
+        diffIm_time = diffIm.getInfo().getVisitInfo().getExposureTime()
```
timj, line 103: Difference images come from visits. Given we can take 2 snaps and there is a gap between the snaps, is the relevant time here the exposure time of a snap or the duration of the visit including the gap (which is greater than the exposure time)?

(also, if `diffim` is an exposure then the docstring for `diffIm` is wrong).

tests/test_association_task.py
```diff
@@ -46,20 +46,30 @@ def setUp(self):
...
+             "diaSourceId": idx + 1 + self.nObjects, "diaObjectId": 0, "trailLength": 5.5*idx}
             for idx in range(self.nSources)])
         self.diaSourceZeroScatter = pd.DataFrame(data=[
             {"ra": 0.04*idx,
              "dec": 0.04*idx,
-             "diaSourceId": idx + 1 + self.nObjects, "diaObjectId": 0}
+             "diaSourceId": idx + 1 + self.nObjects, "diaObjectId": 0, "trailLength": 5.5*idx}
             for idx in range(self.nSources)])
+        visitInfo = afwImage.VisitInfo(exposureTime=30.0,
+                                       )
```
parejkoj, line 57: Why this formatting?

tests/test_association_task.py
```diff
@@ -75,13 +85,36 @@ def test_run(self):
                 [0]):
             self.assertEqual(test_obj_id, expected_obj_id)
 
+    def test_run_trailed_sources(self):
+        """Test the full task by associating a set of diaSources to
+        existing diaObjects when trailed sources are filtered.
+        """
+
+        assocTask = AssociationTask()
+
```
parejkoj, line 94: Unnecessary newlines.

tests/test_association_task.py
```diff
@@ -46,20 +46,30 @@ def setUp(self):
...
         """Test the full task by associating a set of diaSources to
         existing diaObjects.
         """
-        assocTask = AssociationTask()
-        results = assocTask.run(self.diaSources, self.diaObjects)
+
+        config = AssociationTask.ConfigClass()
+        config.doTrailedSourceFilter = False
+        assocTask = AssociationTask(config=config)
+
```
parejkoj, line 71: Unnecessary newlines.

tests/test_association_task.py
```diff
@@ -75,13 +85,36 @@ def test_run(self):
...
+                         len(self.diaObjects) - 3)
+        self.assertEqual(len(results.unAssocDiaSources), 1)
+        for test_obj_id, expected_obj_id in zip(
+                results.matchedDiaSources["diaObjectId"].to_numpy(),
+                [1, 2, 3, 4]):
+            self.assertEqual(test_obj_id, expected_obj_id)
+        for test_obj_id, expected_obj_id in zip(
+                results.unAssocDiaSources["diaObjectId"].to_numpy(),
+                [0]):
+            self.assertEqual(test_obj_id, expected_obj_id)
```
parejkoj, line 109: Why can't you write these like `self.assertEqual(results.matchedDiaSources["diaObjectId"], [1,2,3,4]`? Or at worst use `np.testing.assert_arrays_equal`? The loops make it hard to follow what is being tested.

tests/test_trailedSourceFilter.py
```diff
@@ -0,0 +1,85 @@
+import unittest
```
parejkoj, line 1: Need license block at the top.

tests/test_trailedSourceFilter.py
```diff
@@ -0,0 +1,85 @@
...
+        rng = np.random.default_rng(1234)
+        scatter = 0.1 / 3600
+        self.nSources = 5
+        self.diaSources = pd.DataFrame(data=[
+            {"ra": 0.04*idx + scatter*rng.uniform(-1, 1),
+             "dec": 0.04*idx + scatter*rng.uniform(-1, 1),
+             "diaSourceId": idx, "diaObjectId": 0, "trailLength": 5.5*idx}
+            for idx in range(self.nSources)])
+        visitInfo = afwImage.VisitInfo(exposureTime=30.0,
+                                       )
```
parejkoj, line 23: Why the newline?

tests/test_trailedSourceFilter.py
```diff
@@ -0,0 +1,85 @@
...
+        exposureInfo = afwImage.ExposureInfo()
+        exposureInfo.setVisitInfo(visitInfo)
+        maskedImage = afwImage.MaskedImageF(lsst.geom.Extent2I(64, 64))
+        self.exposure = afwImage.ExposureF(maskedImage, exposureInfo)
+
+    def test_run(self):
+        # Run with default trail length
+
+        trailedSourceFilterTask = TrailedSourceFilterTask()
+
```
parejkoj, line 33: extra newlines

tests/test_trailedSourceFilter.py
```diff
@@ -0,0 +1,85 @@
...
+            for idx in range(self.nSources)])
+        visitInfo = afwImage.VisitInfo(exposureTime=30.0,
+                                       )
+        exposureInfo = afwImage.ExposureInfo()
+        exposureInfo.setVisitInfo(visitInfo)
+        maskedImage = afwImage.MaskedImageF(lsst.geom.Extent2I(64, 64))
+        self.exposure = afwImage.ExposureF(maskedImage, exposureInfo)
+
+    def test_run(self):
+        # Run with default trail length
```
parejkoj, line 30: I think you mean for this to be a docstring for this test method?

tests/test_trailedSourceFilter.py
```diff
@@ -0,0 +1,85 @@
...
+
+class TestTrailedSourceFilterTask(unittest.TestCase):
+
+    def setUp(self):
+        """Create sets of diaSources and diaObjects.
+        """
+        rng = np.random.default_rng(1234)
+        scatter = 0.1 / 3600
+        self.nSources = 5
+        self.diaSources = pd.DataFrame(data=[
```
parejkoj, line 17: Please include a note here about how many long trailed sources this test will generate: it's not at all obvious from the `5.5*idx` approach which of them will be long enough to be filtered.

tests/test_trailedSourceFilter.py
```diff
@@ -0,0 +1,85 @@
...
+
+        trailedSourceFilterTask = TrailedSourceFilterTask()
+
+        results = trailedSourceFilterTask.run(self.diaSources, self.exposure)
+
+        self.assertEqual(len(results.diaSources),
+                         3)
+
+    def test_run_short_max_trail(self):
+        # Run with low trail rate cutoff to test config settings.
```
parejkoj, line 40: For each of these methods, this should be a docstring, not a comment.

tests/test_trailedSourceFilter.py
```diff
@@ -0,0 +1,85 @@
...
+        self.exposure = afwImage.ExposureF(maskedImage, exposureInfo)
+
+    def test_run(self):
+        # Run with default trail length
+
+        trailedSourceFilterTask = TrailedSourceFilterTask()
+
+        results = trailedSourceFilterTask.run(self.diaSources, self.exposure)
+
+        self.assertEqual(len(results.diaSources),
```
parejkoj, line 36: Each of these tests should also have an assert on the contents of the other array returned in the struct (the sources that were filtered). Better to test on which ones were included than just the length: I think this is the first 3 in the first list, and the last two in the second? Similarly in `test_run_short_max_trail`.

tests/test_trailedSourceFilter.py
```diff
@@ -0,0 +1,85 @@
...
+                         0)
+
+    def test_check_dia_source_trail(self):
+        # Check that the task outputs the correct masks.
+
+        trailedSourceFilterTask = TrailedSourceFilterTask()
+        mask = trailedSourceFilterTask.check_dia_source_trail(self.diaSources, self.exposure)
+
+        comparison = mask == [False, False, False, True, True]
+        self.assertTrue(comparison.all())
```
parejkoj, line 72: This makes for a more useful comparison:

```suggestion
        np.testing.assert_array_equal(mask, [False, False, False, True, True])
```

parejkoj (changes requested):
It looks like you marked some comments resolved without making the requested changes (e.g. excessive newlines, only needing the exposure time not the full exposure), and didn't incorporate some others (e.g. referring to the default value in docstrings, making the `check_dia_source_trail` method private)? I've flagged some of them with eyes, but please check for others that were missed.

python/lsst/ap/association/trailedSourceFilter.py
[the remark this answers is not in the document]
parejkoj: I've filed [DM-40769](https://jira.lsstcorp.org/browse/DM-40769) to explore this question. We probably should be doing this inside a context manager in `run`, if we actually want to do it at all. Don't worry about it here; we'll sort it out on that ticket.

python/lsst/ap/association/trailedSourceFilter.py
bsmartradio: Swapped to your wording.

python/lsst/ap/association/trailedSourceFilter.py
bsmartradio: Swapped to your wording.

python/lsst/ap/association/trailedSourceFilter.py
bsmartradio: Updated with maxTrailLength

python/lsst/ap/association/association.py
bsmartradio: Changed doc string to your wording.

python/lsst/ap/association/association.py
bsmartradio: Changed doc string to suggested

python/lsst/ap/association/trailedSourceFilter.py
bsmartradio: Added more to both the opening sentence and the body of the docstring.

python/lsst/ap/association/trailedSourceFilter.py
bsmartradio: Changed it to your suggestion for accfuracy. I also altered the docstring with both the more accurate value and the rough value.

tests/test_trailedSourceFilter.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def test_run(self):
+        # Run with default trail length
+
+        trailedSourceFilterTask = TrailedSourceFilterTask()
+
+        results = trailedSourceFilterTask.run(self.diaSources, self.exposure)
+
+        self.assertEqual(len(results.diaSources),
+                         3)
```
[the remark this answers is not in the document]
bsmartradio: Annoyingly, I had fixed these locally but failed to push the fix. My IDE updated and went a tad over-aggressive on the auto-formatting. I'll make sure to double check.

python/lsst/ap/association/trailedSourceFilter.py
bsmartradio: As a note, I changed it to 36000/2600/24 as the rate needs to be in arcseconds, not degrees.

python/lsst/ap/association/trailedSourceFilter.py
kfindeisen: Is there a policy and/or actual code that defines what the "exposure time" of a snap-combined image is? [The docs for `VisitInfo`](http://doxygen.lsst.codes/stack/doxygen/x_mainDoxyDoc/classlsst_1_1afw_1_1image_1_1_visit_info.html#ab1318e25db6b09d117a8dc78311bff5c) just say:
> get exposure duration (shutter open time); (sec)

Since we can't yet run `CharacterizeImageTask` or `CalibrateImageTask` on multi-snap visits, I'm pretty sure that this case is untestable either way.

timj: At the moment it's defined to be the sum of the exposure time of the snaps. Unlike ObservationInfo, VisitInfo doesn't record the start and end time, only the midpoint (which for two snaps might be a time where no data are being taken).

tests/test_association_task.py
bsmartradio: This results in `ValueError: The truth value of an array with more than one element is ambiguous.
Use a.any() or a.all()` for the arrays, and the original person who wrote the tests I followed likely did it in this way to get around this problem. I looked up ways around this, and this popped up as one of the suggested ways of comparing values in an array for unit testing. The other option seems to be `self.assertTrue(np.array_equal(results.matchedDiaSources["diaObjectId"].values, [1,2], equal_nan=True))`, which I've swapped to since it seems a tad more explicit in what its doing.

tests/test_trailedSourceFilter.py
bsmartradio: Yes, I've cleaned up the comments and made docstrings for each test explaining the test.

tests/test_trailedSourceFilter.py
bsmartradio: Updated the setUp doc string to explicitly list the trail lengths in arcseconds.

tests/test_trailedSourceFilter.py
bsmartradio: Swapped to your suggested code.

tests/test_association_task.py
bsmartradio: Removed

tests/test_association_task.py
parejkoj: Use `np.testing.assert_arrays_equal` then, instead. That will give explicit information which values mismatch. These cannot be NaN, so there's no need for NaN-safe comparisons.

python/lsst/ap/association/association.py
bsmartradio: Removed extra lines

tests/test_association_task.py
bsmartradio: On that note, would it add clarity if I changed the other unit tests to follow this format?

parejkoj: That would definitely be helpful, yes! Please do it on a separate commit, though.

tests/test_trailedSourceFilter.py
bsmartradio: Added tests similar to `self.assertTrue(np.array_equal(results.matchedDiaSources["diaObjectId"].values, [1,2], equal_nan=True))` to check that the output arrays are what is expected.

tests/test_association_task.py
bsmartradio: Swapped to np.testing and also made a separate commit updating the prior unit tests

tests/test_trailedSourceFilter.py
bsmartradio: Updated to np.testing.

python/lsst/ap/association/association.py
```diff
@@ -98,7 +121,16 @@ def run(self,
                 nUpdatedDiaObjects=0,
                 nUnassociatedDiaObjects=0)
 
-        matchResult = self.associate_sources(diaObjects, diaSources)
+        if self.config.doTrailedSourceFilter:
+            diaTrailedResult = self.trailedSourceFilter.run(diaSources, exposure_time)
+            matchResult = self.associate_sources(diaObjects, diaTrailedResult.diaSources)
+
+            self.log.warning("%i DIASources exceed maxTrailLength, dropping "
```
parejkoj, line 128: Why a warning? Nothing went wrong, since the intent was to remove sources. `log.info` is probably fine.

python/lsst/ap/association/diaPipe.py
```diff
@@ -367,8 +367,8 @@ def run(self,
         loaderResult = self.diaCatalogLoader.run(diffIm, self.apdb)
 
         # Associate new DiaSources with existing DiaObjects.
-        assocResults = self.associator.run(diaSourceTable,
-                                           loaderResult.diaObjects)
+        assocResults = self.associator.run(diaSourceTable, loaderResult.diaObjects,
+                                           exposure_time=diffIm.getInfo().getVisitInfo().getExposureTime())
```
parejkoj, line 371: Use the properties:

```suggestion
                                           exposure_time=diffIm.visitInfo.exposureTime)
```

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,112 @@
...
+import lsst.pex.config as pexConfig
+import lsst.pipe.base as pipeBase
+from lsst.utils.timer import timeMethod
+
+
+class TrailedSourceFilterConfig(pexConfig.Config):
+    """Config class for TrailedSourceFilterTask.
+    """
+
+    maxTrailLength = pexConfig.Field(
```
parejkoj, line 33: I guess I missed this earlier: if we're going with snake_case throughout, we should make the configs also snake_case.

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -0,0 +1,112 @@
...
+
+        return pipeBase.Struct(
+            diaSources=dia_sources[~trail_mask].reset_index(drop=True),
+            trailedDiaSources=dia_sources[trail_mask].reset_index(drop=True))
+
+    def _check_dia_source_trail(self, dia_sources, exposure_time):
+        """Find DiaSources that have long trails.
+
+        Creates a mask for sources with lengths greater than 0.416
+        arcseconds/second multiplied by the exposure time.
```
parejkoj, line 93: Again, don't mention the default value, since it's configurable.

```suggestion
        Return a mask of sources with lengths greater than ``config.maxTrailLength``  multiplied by the exposure time.
```

parejkoj (approved):
A few more small comments, and I unresolved and put eyes on a few others that it looks like you missed. Clean these up, and you're good to go.

tests/test_trailedSourceFilter.py
bsmartradio: Missed this one, fixed.

tests/test_trailedSourceFilter.py
bsmartradio: Not relevant anymore as visitInfo has been removed.

python/lsst/ap/association/trailedSourceFilter.py
bsmartradio: Ah yes, I fixed in one place but not another. I've fixed them all now.

python/lsst/ap/association/trailedSourceFilter.py
bsmartradio: Swapped the wording to what you've suggested.

python/lsst/ap/association/association.py
bsmartradio: I saw that the other log.info that drops a source was a warning. However, you are right since dropping the source makes it work as intended so it should be a log.info.

python/lsst/ap/association/trailedSourceFilter.py
bsmartradio: I've swapped to snake case in just trailedSourceFilter.py and left it camel in association.py

python/lsst/ap/association/diaPipe.py
bsmartradio: Swapped to using properties.

bsmartradio:
Jenkins run with this passing: https://rubin-ci.slac.stanford.edu/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/354/pipeline

PR #179, opened 2023-08-29
DM-39938: Allow schema file to be a general URI

timj:
This allows it to be read from anywhere supported by ResourcePath.

Requires lsst/alert_packet#35

### DIFF
--- a/python/lsst/ap/association/packageAlerts.py
+++ b/python/lsst/ap/association/packageAlerts.py
@@ -48,8 +48,8 @@ class PackageAlertsConfig(pexConfig.Config):
     """
     schemaFile = pexConfig.Field(
         dtype=str,
-        doc="Schema definition file for the avro alerts.",
-        default=alertPack.get_path_to_latest_schema()
+        doc="Schema definition file URI for the avro alerts.",
+        default=str(alertPack.get_uri_to_latest_schema())
     )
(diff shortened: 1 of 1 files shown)


python/lsst/ap/association/packageAlerts.py
```diff
@@ -75,7 +76,9 @@ class PackageAlertsTask(pipeBase.Task):
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
-        self.alertSchema = alertPack.Schema.from_file(self.config.schemaFile)
+        uri = ResourcePath(self.config.schemaFile)
+        with uri.as_local() as schemaFile:
+            self.alertSchema = alertPack.Schema.from_file(schemaFile.ospath)
```
ebellm, line 81: Would it make more sense to add a `from_uri` method to `alert_packet.Schema`?  That would require making `alert_packet` dependent on `lsst.resources` but would hide some of this from users.

timj: Yes, if I did that then the lsst.alert.packet.schema could make use of ResourcePath directly and would be a bit simpler internally. I'm happy to switch over to that since it was the more natural thing for me to do, but I was worried you wouldn't like the new dependency (although it is on PyPI so shouldn't be a problem).

I will redo the PRs using ResourcePath to see what you think.

PR #180, opened 2023-10-10
DM-40412: Add edge flag filtering to trailedSourceFilter.py

bsmartradio:
Now that trailed sources which cross over into edge regions are flagged, add in a filter to trailedSourceFilter.py. Currently, this filter is far more aggressive than it should be as it counts trailed sources crossing over into nodata regions as edges. This will be fixed in a later commit.

### DIFF
--- a/python/lsst/ap/association/association.py
+++ b/python/lsst/ap/association/association.py
@@ -112,24 +112,32 @@ class AssociationTask(pipeBase.Task):
               matched to new DiaSources. (`int`)
             - ``nUnassociatedDiaObjects`` : Number of DiaObjects that were
               not matched a new DiaSource. (`int`)
+            - ``longTrailedSources`` : DiaSources which have trail lengths
+              greater than max_trail_length/second*exposure_time.
+              (`pandas.DataFrame``)
         """
         diaSources = self.check_dia_source_radec(diaSources)
+
+        if self.config.doTrailedSourceFilter:
+            diaTrailedResult = self.trailedSourceFilter.run(diaSources, exposure_time)
+            diaSources = diaTrailedResult.diaSources
+            longTrailedSources = diaTrailedResult.longTrailedDiaSources
+
+            self.log.info("%i DiaSources exceed max_trail_length, dropping from source "
+                          "catalog." % len(diaTrailedResult.longTrailedDiaSources))
+            self.metadata.add("num_filtered", len(diaTrailedResult.longTrailedDiaSources))
+        else:
+            longTrailedSources = pd.DataFrame(columns=diaSources.columns)
+
         if len(diaObjects) == 0:
             return pipeBase.Struct(
                 matchedDiaSources=pd.DataFrame(columns=diaSources.columns),
                 unAssocDiaSources=diaSources,
                 nUpdatedDiaObjects=0,
-                nUnassociatedDiaObjects=0)
+                nUnassociatedDiaObjects=0,
+                longTrailedSources=longTrailedSources)
 
-        if self.config.doTrailedSourceFilter:
-            diaTrailedResult = self.trailedSourceFilter.run(diaSources, exposure_time)
-            matchResult = self.associate_sources(diaObjects, diaTrailedResult.diaSources)
-
-            self.log.info("%i DIASources exceed max_trail_length, dropping "
-                          "from source catalog." % len(diaTrailedResult.trailedDiaSources))
-
-        else:
-            matchResult = self.associate_sources(diaObjects, diaSources)
+        matchResult = self.associate_sources(diaObjects, diaSources)
 
         mask = matchResult.diaSources["diaObjectId"] != 0
 
@@ -137,7 +145,8 @@ class AssociationTask(pipeBase.Task):
             matchedDiaSources=matchResult.diaSources[mask].reset_index(drop=True),
             unAssocDiaSources=matchResult.diaSources[~mask].reset_index(drop=True),
             nUpdatedDiaObjects=matchResult.nUpdatedDiaObjects,
-            nUnassociatedDiaObjects=matchResult.nUnassociatedDiaObjects)
+            nUnassociatedDiaObjects=matchResult.nUnassociatedDiaObjects,
+            longTrailedSources=longTrailedSources)
 
     def check_dia_source_radec(self, dia_sources):
         """Check that all DiaSources have non-NaN values for RA/DEC.
--- a/python/lsst/ap/association/trailedSourceFilter.py
+++ b/python/lsst/ap/association/trailedSourceFilter.py
@@ -77,11 +77,11 @@ class TrailedSourceFilterTask(pipeBase.Task):
         result : `lsst.pipe.base.Struct`
             Results struct with components.
 
-            - ``dia_sources`` : DIASource table that is free from unwanted
+            - ``diaSources`` : DIASource table that is free from unwanted
              trailed sources. (`pandas.DataFrame`)
 
-            - ``trailed_dia_sources`` : DIASources that have trails which
-            exceed max_trail_length/second*exposure_time.
+            - ``longTrailedDiaSources`` : DIASources that have trails which
+            exceed max_trail_length/second*exposure_time (seconds).
             (`pandas.DataFrame`)
         """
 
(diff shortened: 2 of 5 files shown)


python/lsst/ap/association/association.py
```diff
@@ -114,30 +114,39 @@ def run(self,
...
         diaSources = self.check_dia_source_radec(diaSources)
+
+        if self.config.doTrailedSourceFilter:
+            diaTrailedResult = self.trailedSourceFilter.run(diaSources,
+                                                            exposure_time)
+            diaSources = diaTrailedResult.diaSources
+            trailedSources = diaTrailedResult.trailedDiaSources
+
+            self.log.info(
+                "%i DIASources exceed max_trail_length, dropping fromsource catalog." % len(
```
mrawls, line 125:
```suggestion
                "%I Found one or more DiaSource that exceeds max_trail_length, dropping from source catalog." % len(
```
It might be even more informative if "num_filtered" were part of this log message.

python/lsst/ap/association/association.py
```diff
@@ -114,30 +114,39 @@ def run(self,
               not matched a new DiaSource. (`int`)
```
mrawls, line 114: Please update the docstring to describe the additional things that are returned now.

python/lsst/ap/association/association.py
```diff
@@ -114,30 +114,39 @@ def run(self,
...
-        if self.config.doTrailedSourceFilter:
-            diaTrailedResult = self.trailedSourceFilter.run(diaSources, exposure_time)
-            matchResult = self.associate_sources(diaObjects, diaTrailedResult.diaSources)
-
-            self.log.info("%i DIASources exceed max_trail_length, dropping "
-                          "from source catalog." % len(diaTrailedResult.trailedDiaSources))
-
-        else:
-            matchResult = self.associate_sources(diaObjects, diaSources)
+        matchResult = self.associate_sources(diaObjects, diaSources)
```
mrawls, line 140: Was this implemented wrong before and your commit is fixing a bug? Just trying to understand

python/lsst/ap/association/association.py
```diff
@@ -114,30 +114,39 @@ def run(self,
...
+            diaSources = diaTrailedResult.diaSources
+            trailedSources = diaTrailedResult.trailedDiaSources
+
+            self.log.info(
+                "%i DIASources exceed max_trail_length, dropping fromsource catalog." % len(
+                    diaTrailedResult.trailedDiaSources))
+            self.metadata.add("num_filtered",
+                              len(diaTrailedResult.trailedDiaSources))
+        else:
+            trailedSources = pd.DataFrame(columns=diaSources.columns)
```
mrawls, line 130: Consider a comment along the lines of, "Create an empty trailedSources df to add to the return Struct." It's a little counterintuitive to me that "trailedSources" is only populated if we decide to drop long trailed sources. Makes me wonder what happens to any short trailed sources...

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -79,32 +84,42 @@ def run(self, dia_sources, exposure_time):
...
-        ``config.max_trail_length``  multiplied by the exposure time.
+        ``config.max_trail_length``  multiplied by the exposure time and
+        have ext_trailedSources_Naive_flag_edge set.
 
         Parameters
         ----------
         dia_sources : `pandas.DataFrame`
             Input DIASources to check for trail lengths.
         exposure_time : `float`
             Exposure time from difference image.
```
mrawls, line 110: Presumably in seconds

python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -79,32 +84,42 @@ def run(self, dia_sources, exposure_time):
...
             trailedDiaSources=dia_sources[trail_mask].reset_index(drop=True))
 
-    def _check_dia_source_trail(self, dia_sources, exposure_time):
+    def _check_dia_source_trail(self, dia_sources, exposure_time, flags):
         """Find DiaSources that have long trails.
 
         Return a mask of sources with lengths greater than
-        ``config.max_trail_length``  multiplied by the exposure time.
+        ``config.max_trail_length``  multiplied by the exposure time and
+        have ext_trailedSources_Naive_flag_edge set.
```
mrawls, line 103: Why is this an "and" and not an "or" situation?

tests/test_trailedSourceFilter.py
```diff
@@ -94,15 +110,43 @@ def test_run_no_trails(self):
...
+        """
+        trailedSourceFilterTask = TrailedSourceFilterTask()
+
+        results = trailedSourceFilterTask.run(self.edgeDiaSources, self.exposure_time)
+
+        self.assertEqual(len(results.diaSources), 3)
+        np.testing.assert_array_equal(results.diaSources['diaSourceId'].values, [0, 2, 3])
+        np.testing.assert_array_equal(results.trailedDiaSources['diaSourceId'].values, [1, 4])
+
     def test_check_dia_source_trail(self):
```
mrawls, line 125: Please consider a clearer test name

tests/test_trailedSourceFilter.py
```diff
@@ -94,15 +110,43 @@ def test_run_no_trails(self):
...
 
         Test that the mask filter returns the expected mask array.
         """
         trailedSourceFilterTask = TrailedSourceFilterTask()
-        mask = trailedSourceFilterTask._check_dia_source_trail(self.diaSources, self.exposure_time)
+        flag_map = os.path.join(utils.getPackageDir("ap_association"), "data/association-flag-map.yaml")
+        unpacker = UnpackApdbFlags(flag_map, "DiaSource")
+        flags = unpacker.unpack(self.diaSources["flags"], "flags")
+        mask = trailedSourceFilterTask._check_dia_source_trail(self.diaSources, self.exposure_time,
+                                                               flags)
```
mrawls, line 135: Please consider a different variable name, since "mask" means something pretty specific in Science Pipelines parlance and I'm pretty sure it is not this :)

mrawls (commented):
My suggestions are all to enhance clarify, not to change any functionality, generally this looks good and I'm glad it is working now!

python/lsst/ap/association/association.py
bsmartradio: num_filtered is the same thing as the log message, they are both len(diaTrailedResult.trailedDiaSources). do you mean I should do something like `from source catalog. num_filtered: " % len( `

python/lsst/ap/association/association.py
bsmartradio: Added

python/lsst/ap/association/association.py
bsmartradio: This wasn't implemented wrong per say, but trailed sources would not be run on a diaSource catalog if there were no diaObjects. This meant there were situation where the filtering was skipped, so I moved things around so it would always run and the object check would be on an already filtered diaSource catalog.

python/lsst/ap/association/association.py
bsmartradio: The intention of trailed sources is just to contain the sources that will be dropped, however we may want to think of renaming it?

python/lsst/ap/association/association.py
mrawls: OH sorry it's a super long line isn't it, hello that's what the `%i` is, my bad!

python/lsst/ap/association/association.py
mrawls: I think I'm partial to `longTrailedSources` throughout, if that's not too much of a headache. (If they go off the edge, odds are good they are long, so no need to distinguish between edge overlapping trailed sources vs long trailed sources that may or may not intersect an edge IMO)

python/lsst/ap/association/trailedSourceFilter.py
bsmartradio: max_trail_length is arcseconds/second, so the output is arcseconds. Should I clarify that there? I have it defined in the description of the task and in the pexConfig documentation.

python/lsst/ap/association/trailedSourceFilter.py
bsmartradio: Your right, its an or, possibly an and/or situation. Updated it to or.

tests/test_trailedSourceFilter.py
bsmartradio: Which test are you suggesting a clearer name for? `test_check_dia_source_trail` is checking the `_check_dia_source_trail` function in trailedSourceFilter. I can rename that if you think that needs a clearer name.

tests/test_trailedSourceFilter.py
bsmartradio: I can swap it to trailed_mask like I have elsewhere, though maybe I should rename the whole lot to not imply they are the different kind of masks...

python/lsst/ap/association/trailedSourceFilter.py
mrawls: I meant that while you're editing this method, explicitly defining the `exposure_time` parameter as having units of seconds would be helpful

tests/test_trailedSourceFilter.py
mrawls: Yeah, I meant the one on this line, `test_check_dia_source_trail` - it felt like "test" and "check" were weirdly doubled but I see now why that is the name you chose! Hmm. Maybe just change the description, e.g., "Test that the DiaSource trail checker is correctly identifying long trails" ?

I'm realizing a lot of my uncertainty about what is happening is connected to the use of terms "filter", "mask", "flag", "check", "sort" semi-interchangeably. The renaming here to `trailed_source_mask` helps!

python/lsst/ap/association/trailedSourceFilter.py
bsmartradio: Oooh, yes! It wasn't highlighted so I wasn't sure what you were referencing. Added.

PR #181, opened 2023-10-18
DM-32769: Implement streak detection and masking in single frame processing

mrawls:
This PR adds two flags to the flag map. One is so the new streak mask info makes it into `associatedDiaSources`. The other is for backwards compatibility with DM-40412.


bsmartradio (approved):
Look good, yay new flags!

PR #182, opened 2023-10-18
DM-41024: Add trailed source metric

bsmartradio:


python/lsst/ap/association/diaPipe.py
```diff
@@ -114,6 +114,12 @@ class DiaPipelineConnections(
         storageClass="DataFrame",
         dimensions=("instrument", "visit", "detector"),
     )
+    longTrailedSources = pipeBase.connectionTypes.Output(
+        doc="Optional output temporarily storing long trailed diaSources.",
+        dimensions=("instrument", "visit", "detector"),
+        storageClass="DataFrame",
+        name="{fakesType}{coaddName}Diff_longTrailedSrc",
```
mrawls, line 121: Could we call it `longTrailedSources` or are you partial to `Src`?

I guess this does parallel `assocDiaSrc` and `diaForcedSrc` etc. so you are just following that... makes me think it should be `longTrlSrc` (don't actually do that)

mrawls (approved):

python/lsst/ap/association/diaPipe.py
bsmartradio: I think I will stick with Src only to mirror the others, but I agree shortening it would be a bad idea.

PR #188, opened 2024-01-04
DM-42354: Populate diaForcedSource flags with zeros

ebellm:


python/lsst/ap/association/diaForcedSource.py
```diff
@@ -235,6 +235,8 @@ def _calibrate_and_merge(self,
         output_catalog["midpointMjdTai"] = midpointMjdTai
         output_catalog["band"] = diff_exp.getFilter().bandLabel
         output_catalog["time_processed"] = DateTime.now().toPython()
```
kfindeisen, line 237: Pre-existing, but should this field be set by `ForcedSourceTask`?

kfindeisen (approved):
Looks good! Just one unrelated question about placeholders.

PR #189, opened 2024-01-14
DM-39466: Added flagging of sources near the INJECTED and INJECTED_TEMPLATE 

BrunoSanchez:
Added flagging of sources near the INJECTED and INJECTED_TEMPLATE  mask planes


BrunoSanchez:
Same here. 
I didn't assign the correct reviewers

isullivan (approved):

PR #190, opened 2024-01-18
DM-40798: Exclude sky sources from RB classification

rai-harshit:
Create a new pipelineTask that selects sources from the diaSource catalog and writes a goodSeeingDiff_candidateDiaSrc (or similar name) catalog that can be used as an input to TransiNetTask.

### DIFF
--- a/python/lsst/ap/association/__init__.py
+++ b/python/lsst/ap/association/__init__.py
@@ -19,11 +19,12 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-from .version import *
-from .trailedSourceFilter import *
 from .association import *
 from .diaForcedSource import *
+from .filterDiaSourceCatalog import *
 from .loadDiaCatalogs import *
 from .packageAlerts import *
 from .diaPipe import *
+from .trailedSourceFilter import *
 from .transformDiaSourceCatalog import *
+from .version import *
--- a/python/lsst/ap/association/filterDiaSourceCatalog.py
+++ b/python/lsst/ap/association/filterDiaSourceCatalog.py
@@ -1,0 +1,125 @@
+# This file is part of ap_association
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+__all__ = (
+    "FilterDiaSourceCatalogConfig",
+    "FilterDiaSourceCatalogTask",
+)
+
+import numpy as np
+
+from lsst.afw.table import SourceCatalog
+import lsst.pex.config as pexConfig
+import lsst.pipe.base as pipeBase
+import lsst.pipe.base.connectionTypes as connTypes
+from lsst.utils.timer import timeMethod
+
+
+class FilterDiaSourceCatalogConnections(
+    pipeBase.PipelineTaskConnections,
+    dimensions=("instrument", "visit", "detector"),
+    defaultTemplates={"coaddName": "deep", "fakesType": ""},
+):
+    """Connections class for FilterDiaSourceCatalogTask."""
+
+    diaSourceCat = connTypes.Input(
+        doc="Catalog of DiaSources produced during image differencing.",
+        name="{fakesType}{coaddName}Diff_diaSrc",
+        storageClass="SourceCatalog",
+        dimensions=("instrument", "visit", "detector"),
+    )
+
+    filteredDiaSourceCat = connTypes.Output(
+        doc="Output catalog of DiaSources after filtering.",
+        name="{fakesType}{coaddName}Diff_candidateDiaSrc",
+        storageClass="SourceCatalog",
+        dimensions=("instrument", "visit", "detector"),
+    )
+
+    rejectedDiaSources = connTypes.Output(
+        doc="Optional output storing all the rejected DiaSources.",
+        name="{fakesType}{coaddName}Diff_rejectedDiaSrc",
+        storageClass="SourceCatalog",
+        dimensions={"instrument", "visit", "detector"},
+    )
+
+    def __init__(self, *, config=None):
+        super().__init__(config=config)
+        if not self.config.doWriteRejectedSources:
+            self.outputs.remove("rejectedDiaSources")
+
+
+class FilterDiaSourceCatalogConfig(
+    pipeBase.PipelineTaskConfig, pipelineConnections=FilterDiaSourceCatalogConnections
+):
+    """Config class for FilterDiaSourceCatalogTask."""
+
+    doRemoveSkySources = pexConfig.Field(
+        dtype=bool,
+        default=False,
+        doc="Input DiaSource catalog contains SkySources that should be "
+        "removed before storing the output DiaSource catalog.",
+    )
+
+    doWriteRejectedSources = pexConfig.Field(
+        dtype=bool,
+        default=True,
+        doc="Store the output DiaSource catalog containing all the rejected "
+        "sky sources."
+    )
+
+
+class FilterDiaSourceCatalogTask(pipeBase.PipelineTask):
+    """Filter out sky sources from a DiaSource catalog."""
+
+    ConfigClass = FilterDiaSourceCatalogConfig
+    _DefaultName = "filterDiaSourceCatalog"
+
+    @timeMethod
+    def run(self, diaSourceCat):
+        """Filter sky sources from the supplied DiaSource catalog.
+
+        Parameters
+        ----------
+        diaSourceCat : `lsst.afw.table.SourceCatalog`
+            Catalog of sources measured on the difference image.
+
+        Returns
+        -------
+        filterResults : `lsst.pipe.base.Struct`
+
+            ``filteredDiaSourceCat`` : `lsst.afw.table.SourceCatalog`
+                The catalog of filtered sources.
+            ``rejectedDiaSources`` : `lsst.afw.table.SourceCatalog`
+                The catalog of rejected sources.
+        """
+        rejectedSkySources = None
+        if self.config.doRemoveSkySources:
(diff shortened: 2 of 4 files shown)


python/lsst/ap/association/filterDiaSourceCatalog.py
```diff
@@ -0,0 +1,107 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+__all__ = (
+    "FilterDiaSourceCatalogConnections",
```
isullivan, line 23: We shouldn't need the connections class in `__all__`

python/lsst/ap/association/filterDiaSourceCatalog.py
```diff
@@ -0,0 +1,107 @@
...
+        inputs = butlerQC.get(inputRefs)
+        idGenerator = self.config.idGenerator.apply(butlerQC.quantum.dataId)
+        inputs["ccdVisitId"] = idGenerator.catalog_id
+
+        outputs = self.run(**inputs)
+
+        butlerQC.put(outputs, outputRefs)
+
+    @timeMethod
+    def run(self, diaSourceCat, ccdVisitId):
```
isullivan, line 89: `ccdVisitId` is not actually used here, I suggest removing it.

python/lsst/ap/association/filterDiaSourceCatalog.py
```diff
@@ -0,0 +1,107 @@
...
+    _DefaultName = "filterDiaSourceCatalog"
+
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
+        inputs = butlerQC.get(inputRefs)
+        idGenerator = self.config.idGenerator.apply(butlerQC.quantum.dataId)
+        inputs["ccdVisitId"] = idGenerator.catalog_id
+
+        outputs = self.run(**inputs)
+
+        butlerQC.put(outputs, outputRefs)
```
isullivan, line 86: If you remove `ccdVisitId` below, you can remove the entire `runQuantum` method.

python/lsst/ap/association/filterDiaSourceCatalog.py
```diff
@@ -0,0 +1,107 @@
...
+    """Config class for FilterDiaSourceCatalogTask."""
+
+    doRemoveSkySources = pexConfig.Field(
+        dtype=bool,
+        default=False,
+        doc="Input DiaSource catalog contains SkySources that should be "
+        "removed before storing the output DiaSource catalog.",
+    )
+
+    idGenerator = DetectorVisitIdGeneratorConfig.make_field()
```
isullivan, line 70: If you remove `ccdVisitId` below, then you can also remove `idGenerator`.

python/lsst/ap/association/filterDiaSourceCatalog.py
```diff
@@ -0,0 +1,107 @@
...
+
+        Returns
+        -------
+        results : `lsst.pipe.base.Struct`
+            Results struct with components.
+        """
+        if self.config.doRemoveSkySources:
+            sky_source_column = diaSourceCat["sky_source"]
+            num_sky_sources = np.sum(sky_source_column)
+            diaSourceCat = diaSourceCat[~sky_source_column]
```
isullivan, line 105: This will result in a non-contiguous catalog, which may break downstream code. Use `diaSourceCat = diaSourceCat[~sky_source_column].copy(deep=True)` to return a contiguous catalog.

python/lsst/ap/association/filterDiaSourceCatalog.py
```diff
@@ -0,0 +1,107 @@
...
+        storageClass="SourceCatalog",
+        dimensions=("instrument", "visit", "detector"),
+    )
+
+    filteredDiaSourceCat = connTypes.Output(
+        doc="Output catalog of DiaSources after filtering.",
+        name="{fakesType}{coaddName}Diff_candidateDiaSrc",
+        storageClass="SourceCatalog",
+        dimensions=("instrument", "visit", "detector"),
+    )
```
isullivan, line 55: Please add an optional output of all of the filtered diaSources. You can make an output optional by adding the connection here as normal and adding a config option below such as `doWriteSkySources` with a default of True. Then follow [this example](https://github.com/lsst/ap_association/blob/6b1805f162f415617d71cbf0f2a1a5217842fb36/python/lsst/ap/association/diaPipe.py#L135) to remove the connection if the config is turned off.

tests/test_filterDiaSourceCatalog.py
```diff
@@ -0,0 +1,71 @@
...
+        self.expId = 4321
+        self.bbox = geom.Box2I(geom.Point2I(0, 0),
+                               geom.Extent2I(1024, 1153))
+        dataset = measTests.TestDataset(self.bbox)
+        for srcIdx in range(self.nSources):
+            dataset.addSource(10000.0, geom.Point2D(srcIdx, self.yLoc))
+        schema = dataset.makeMinimalSchema()
+        schema.addField("sky_source", type="Flag", doc="Sky objects.")
+        _, self.diaSourceCat = dataset.realize(10.0, schema, randomSeed=1234)
+        self.diaSourceCat[0:5]["sky_source"] = True
```
isullivan, line 44: Please make the number of sky sources a class parameter `self.nSkySources = 5` and use that here.

tests/test_filterDiaSourceCatalog.py
```diff
@@ -0,0 +1,71 @@
...
+        self.config.doRemoveSkySources = False
+        filterDiaSourceCatalogTask = FilterDiaSourceCatalogTask(config=self.config)
+        result = filterDiaSourceCatalogTask.run(self.diaSourceCat, ccdVisitId=self.expId)
+        self.assertEqual(len(result.filteredDiaSourceCat), len(self.diaSourceCat))
+
+    def test_run_no_filter(self):
+        self.config.doRemoveSkySources = True
+        filterDiaSourceCatalogTask = FilterDiaSourceCatalogTask(config=self.config)
+        result = filterDiaSourceCatalogTask.run(self.diaSourceCat, ccdVisitId=self.expId)
+        self.assertEqual(len(result.filteredDiaSourceCat),
```
isullivan, line 57: Add an additional couple `assertEqual`s comparing the total number of returned sources to `self.nSources - self.nSkySources`, and the number of sources in the returned (that I'd like you to add) sky sources catalog to `self.nSkySources`

isullivan (changes requested):
I'd like to see fairly substantial changes in this package, so I would like to re-review after you have addressed my current comments.

isullivan (commented):
I forgot to add in my initial review a request to deprecate the config `doRemoveSkySources` in `TransformDiaSourceCatalogTask`, and remove the code that handles sky sources around line 226 of that file. Instead, you could add a check to see if any sources are flagged as sky sources, and either raise an error if any are found.

python/lsst/ap/association/filterDiaSourceCatalog.py
rai-harshit: Sure. I tried to mimic the setup of `transformDiaSourceCatalog.py` which does include the connections class in `__all__`. Is that done for a specific purpose that I'm not aware of?

tests/test_filterDiaSourceCatalog.py
rai-harshit: I've added an assertEqual for your first suggestion. I'm a bit unclear on the second suggestion- mostly confused by the `that I'd like you to add`

isullivan: See my above comment about adding an optional output to your task that returns the catalog of sources that have been removed.

python/lsst/ap/association/filterDiaSourceCatalog.py
```diff
@@ -0,0 +1,117 @@
...
+        ----------
+        diaSourceCat : `lsst.afw.table.SourceCatalog`
+            Catalog of sources measured on the difference image.
+
+        Returns
+        -------
+        results : `lsst.pipe.base.Struct`
+            Results struct with components.
+        """
+        rejectedSkySources = SourceCatalog()
```
isullivan, line 109: I don't think this will work if `doRemoveSkySources` is False, since it won't have the right schema and would probably raise an error if you tried to write it to the Butler. You overwrite it later if we do remove sky sources, so just set it to `None` here and check if it is `None` before the return. If it is, set it to `rejectedSkySources = SourceCatalog(diaSourceCat.getSchema())`.

tests/test_filterDiaSourceCatalog.py
```diff
@@ -0,0 +1,78 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import unittest
```
isullivan, line 22: Leave a blank line after the `unittest` import.

tests/test_filterDiaSourceCatalog.py
```diff
@@ -0,0 +1,78 @@
...
+import lsst.utils.tests
+import lsst.meas.base.tests as measTests
+import lsst.geom as geom
+from lsst.ap.association.filterDiaSourceCatalog import (FilterDiaSourceCatalogConfig,
```
isullivan, line 26: Alphabetize the LSST imports.

tests/test_filterDiaSourceCatalog.py
```diff
@@ -0,0 +1,78 @@
...
+        result = filterDiaSourceCatalogTask.run(self.diaSourceCat)
+        self.assertEqual(len(result.filteredDiaSourceCat), len(self.diaSourceCat))
+        self.assertEqual(len(result.rejectedDiaSources), 0)
+
+    def test_run_with_filter(self):
+        self.config.doRemoveSkySources = True
+        self.config.doWriteSkySources = True
+        filterDiaSourceCatalogTask = FilterDiaSourceCatalogTask(config=self.config)
+        result = filterDiaSourceCatalogTask.run(self.diaSourceCat)
+        nExpectedFilteredSources = self.nSources - self.nSkySources
```
isullivan, line 61: It will pass, but please add a line with `self.assertEqual(len(self.diaSourceCat), self.nSources)` to enforce that `FilterDiaSourceCatalogTask` does not modify the original catalog.

python/lsst/ap/association/__init__.py
```diff
@@ -27,3 +27,4 @@
 from .packageAlerts import *
 from .diaPipe import *
 from .transformDiaSourceCatalog import *
+from .filterDiaSourceCatalog import *
```
isullivan, line 30: We haven't been good about keeping these imports alphabetical, but you could fix that on this ticket.

rai-harshit: I tried re-arranging the imports, but after one point, it started complaining about circular imports. If this needs to be addressed, we can probably have another ticket.

PR #191, opened 2024-02-01
DM-40414: Write PipelineTask to transmit alerts to Kafka

bsmartradio:
Add a producer to packageAlerts in ap_association. The producer will send alerts from the pipeline to the alert stream using several environment variables to set the server, user, password, and topic.

Jenkins run here: https://rubin-ci.slac.stanford.edu/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/571/pipeline

### DIFF
--- a/python/lsst/ap/association/packageAlerts.py
+++ b/python/lsst/ap/association/packageAlerts.py
@@ -24,13 +24,20 @@ __all__ = ("PackageAlertsConfig", "PackageAlertsTask")
 import io
 import os
 import sys
-import logging
 
 from astropy import wcs
 import astropy.units as u
 from astropy.nddata import CCDData, VarianceUncertainty
 import pandas as pd
 import struct
+import fastavro
+# confluent_kafka is not in the standard Rubin environment as it is a third
+# party package and is only needed when producing alerts.
+try:
+    import confluent_kafka
+    from confluent_kafka import KafkaException
+except ImportError:
+    confluent_kafka = None
 
 import lsst.alert.packet as alertPack
 import lsst.afw.geom as afwGeom
@@ -39,13 +46,6 @@ import lsst.pex.config as pexConfig
 from lsst.pex.exceptions import InvalidParameterError
 import lsst.pipe.base as pipeBase
 from lsst.utils.timer import timeMethod
-import fastavro
-
-"""Methods for packaging Apdb and Pipelines data into Avro alerts.
-"""
-
-_log = logging.getLogger("lsst." + __name__)
-_log.setLevel(logging.DEBUG)
 
 
 class PackageAlertsConfig(pexConfig.Config):
@@ -71,14 +71,14 @@ class PackageAlertsConfig(pexConfig.Config):
 
     doProduceAlerts = pexConfig.Field(
         dtype=bool,
-        doc="Turn on alert production to kafka if true. Set to false by default",
+        doc="Turn on alert production to kafka if true and if confluent_kafka is in the environment.",
         default=False,
     )
 
     doWriteAlerts = pexConfig.Field(
         dtype=bool,
-        doc="Write alerts to disk if true. Set to true by default",
-        default=True,
+        doc="Write alerts to disk if true.",
+        default=False,
     )
 
 
@@ -96,51 +96,46 @@ class PackageAlertsTask(pipeBase.Task):
         os.makedirs(self.config.alertWriteLocation, exist_ok=True)
 
         if self.config.doProduceAlerts:
+            if confluent_kafka is not None:
+                self.password = os.getenv("AP_KAFKA_PRODUCER_PASSWORD")
+                if not self.password:
+                    raise ValueError("Kafka password environment variable was not set.")
+                self.username = os.getenv("AP_KAFKA_PRODUCER_USERNAME")
+                if not self.username:
+                    raise ValueError("Kafka username environment variable was not set.")
+                self.server = os.getenv("AP_KAFKA_SERVER")
+                if not self.server:
+                    raise ValueError("Kafka server environment variable was not set.")
+                self.kafkaTopic = os.getenv("AP_KAFKA_TOPIC")
+                if not self.kafkaTopic:
+                    raise ValueError("Kafka topic environment variable was not set.")
 
-            self.password = os.getenv("AP_KAFKA_PRODUCER_PASSWORD")
-            self.username = os.getenv("AP_KAFKA_PRODUCER_USERNAME")
-            self.server = os.getenv("AP_KAFKA_SERVER")
-            self.kafka_topic = os.getenv("AP_KAFKA_TOPIC")
-            # confluent_kafka configures all of its classes with dictionaries. This one
-            # sets up the bare minimum that is needed.
-            self.kafka_config = {
-                # This is the URL to use to connect to the Kafka cluster.
-                "bootstrap.servers": self.server,
-                # These next two properties tell the Kafka client about the specific
-                # authentication and authorization protocols that should be used when
-                # connecting.
-                "security.protocol": "SASL_PLAINTEXT",
-                "sasl.mechanisms": "SCRAM-SHA-512",
-                # The sasl.username and sasl.password are passed through over
-                # SCRAM-SHA-512 auth to connect to the cluster. The username is not
-                # sensitive, but the password is (of course) a secret value which
-                # should never be committed to source code.
-                "sasl.username": self.username,
-                "sasl.password": self.password,
-                # Batch size limits the largest size of a kafka alert that can be sent.
-                # We set the batch size to 2 Mb.
-                "batch.size": 2097152,
-                "linger.ms": 5,
-            }
+                # confluent_kafka configures all of its classes with dictionaries. This one
+                # sets up the bare minimum that is needed.
+                self.kafkaConfig = {
+                    # This is the URL to use to connect to the Kafka cluster.
+                    "bootstrap.servers": self.server,
+                    # These next two properties tell the Kafka client about the specific
+                    # authentication and authorization protocols that should be used when
+                    # connecting.
+                    "security.protocol": "SASL_PLAINTEXT",
+                    "sasl.mechanisms": "SCRAM-SHA-512",
+                    # The sasl.username and sasl.password are passed through over
+                    # SCRAM-SHA-512 auth to connect to the cluster. The username is not
+                    # sensitive, but the password is (of course) a secret value which
+                    # should never be committed to source code.
+                    "sasl.username": self.username,
+                    "sasl.password": self.password,
+                    # Batch size limits the largest size of a kafka alert that can be sent.
+                    # We set the batch size to 2 Mb.
+                    "batch.size": 2097152,
+                    "linger.ms": 5,
+                }
+                self.producer = confluent_kafka.Producer(**self.kafkaConfig)
 
-            try:
-                from confluent_kafka import KafkaException
-                self.kafka_exception = KafkaException
-                import confluent_kafka
-            except ImportError as error:
-                error.add_note("Could not import confluent_kafka. Alerts will not be sent "
-                               "to the alert stream")
-                _log.error(error)
-
-            if not self.password:
-                raise ValueError("Kafka password environment variable was not set.")
-            if not self.username:
-                raise ValueError("Kafka username environment variable was not set.")
-            if not self.server:
-                raise ValueError("Kafka server environment variable was not set.")
-            if not self.kafka_topic:
-                raise ValueError("Kafka topic environment variable was not set.")
-            self.producer = confluent_kafka.Producer(**self.kafka_config)
+            else:
+                raise RuntimeError("Produce alerts is set but confluent_kafka is not present in "
+                                   "the environment. Alerts will not be sent to the alert stream.")
 
     @timeMethod
     def run(self,
@@ -152,6 +147,10 @@ class PackageAlertsTask(pipeBase.Task):
             template,
             ):
         """Package DiaSources/Object and exposure data into Avro alerts.
+
+        Alerts can be sent to the alert stream if ``doProduceAlerts`` is set
+        and written to disk if ``doWriteAlerts`` is set. Both can be set at the
+        same time, and are independent of one another.
 
         Writes Avro alerts to a location determined by the
         ``alertWriteLocation`` configurable.
@@ -223,20 +222,12 @@ class PackageAlertsTask(pipeBase.Task):
                                    diffImCutout,
                                    templateCutout))
 
-        if self.config.doProduceAlerts and "confluent_kafka" in sys.modules:
+        if self.config.doProduceAlerts:
             self.produceAlerts(alerts, ccdVisitId)
 
-        elif self.config.doProduceAlerts and "confluent_kafka" not in sys.modules:
-            raise Exception("Produce alerts is set but confluent_kafka is not in the environment.")
-
         if self.config.doWriteAlerts:
-            with open(os.path.join(self.config.alertWriteLocation,
-                                   f"{ccdVisitId}.avro"),
-                      "wb") as f:
+            with open(os.path.join(self.config.alertWriteLocation, f"{ccdVisitId}.avro"), "wb") as f:
                 self.alertSchema.store_alerts(f, alerts)
-
-        if not self.config.doProduceAlerts and not self.config.doWriteAlerts:
-            raise Exception("Neither produce alerts nor write alerts is set.")
 
     def _patchDiaSources(self, diaSources):
         """Add the ``programId`` column to the data.
@@ -249,7 +240,7 @@ class PackageAlertsTask(pipeBase.Task):
         diaSources["programId"] = 0
 
     def createDiaSourceExtent(self, bboxSize):
-        """Create a extent for a  box for the cutouts given the size of the
+        """Create an extent for a box for the cutouts given the size of the
         square BBox that covers the source footprint.
 
         Parameters
@@ -270,19 +261,29 @@ class PackageAlertsTask(pipeBase.Task):
         return extent
 
     def produceAlerts(self, alerts, ccdVisitId):
+        """Serialize alerts and send them to the alert stream using
+        confluent_kafka's producer.
 
+        Parameters
+        ----------
+        alerts : `dict`
+            Dictionary of alerts to be sent to the alert stream.
+        ccdVisitId : `int`
+            ccdVisitId of the alerts sent to the alert stream. Used to write
+            out alerts which fail to be sent to the alert stream.
+        """
         for alert in alerts:
-            alert_bytes = self._serialize_alert(alert, schema=self.alertSchema.definition, schema_id=1)
+            alertBytes = self._serializeAlert(alert, schema=self.alertSchema.definition, schema_id=1)
             try:
-                self.producer.produce(self.kafka_topic, alert_bytes, callback=self._delivery_callback)
+                self.producer.produce(self.kafkaTopic, alertBytes, callback=self._delivery_callback)
                 self.producer.flush()
 
-            except self.kafka_exception as e:
-                _log.error('Kafka error: {}, message was {} bytes'.format(e, sys.getsizeof(alert_bytes)))
+            except KafkaException as e:
+                self.log.warning('Kafka error: {}, message was {} bytes'.format(e, sys.getsizeof(alertBytes)))
 
                 with open(os.path.join(self.config.alertWriteLocation,
                                        f"{ccdVisitId}_{alert['alertId']}.avro"), "wb") as f:
-                    f.write(alert_bytes)
+                    f.write(alertBytes)
 
         self.producer.flush()
 
@@ -466,11 +467,11 @@ class PackageAlertsTask(pipeBase.Task):
             cutoutBytes = streamer.getvalue()
         return cutoutBytes
 
-    def _serialize_alert(self, alert, schema=None, schema_id=0):
+    def _serializeAlert(self, alert, schema=None, schema_id=0):
         """Serialize an alert to a byte sequence for sending to Kafka.
 
         Parameters
-        ----------`
+        ----------
         alert : `dict`
             An alert payload to be serialized.
         schema : `dict`, optional
@@ -479,7 +480,7 @@ class PackageAlertsTask(pipeBase.Task):
         schema_id : `int`, optional
             The Confluent Schema Registry ID of the schema. By default, 0 (an
             invalid ID) is used, indicating that the schema is not registered.
-    `
+
         Returns
         -------
         serialized : `bytes`
@@ -491,38 +492,12 @@ class PackageAlertsTask(pipeBase.Task):
 
         buf = io.BytesIO()
         # TODO: Use a proper schema versioning system (DM-42606)
-        buf.write(self._serialize_confluent_wire_header(schema_id))
+        buf.write(self._serializeConfluentWireHeader(schema_id))
         fastavro.schemaless_writer(buf, schema, alert)
         return buf.getvalue()
 
-    def _deserialize_alert(self, alert_bytes, schema=None):
-        """Deserialize an alert message from Kafka.
-
-        Paramaters
-        ----------
-        alert_bytes : `bytes`
-            Binary-encoding serialized Avro alert, including Confluent Wire
-            Format prefix.
-        schema : `dict`, optional
-            An Avro schema definition describing how to encode `alert`. By default,
-            the latest schema is used.
-
-        Returns
-        -------
-        alert : `dict`
-            An alert payload.
-        """
-        if schema is None:
-            schema = self.alertSchema.definition
-
-        header_bytes = alert_bytes[:5]
-        version = self._deserialize_confluent_wire_header(header_bytes)
-        assert version == 0
-        content_bytes = io.BytesIO(alert_bytes[5:])
-        return fastavro.schemaless_reader(content_bytes, schema)
-
     @staticmethod
-    def _serialize_confluent_wire_header(schema_version):
+    def _serializeConfluentWireHeader(schema_version):
         """Returns the byte prefix for Confluent Wire Format-style Kafka messages.
 
         Parameters
@@ -545,29 +520,8 @@ class PackageAlertsTask(pipeBase.Task):
         ConfluentWireFormatHeader = struct.Struct(">bi")
         return ConfluentWireFormatHeader.pack(0, schema_version)
 
-    @staticmethod
-    def _deserialize_confluent_wire_header(raw):
-        """Parses the byte prefix for Confluent Wire Format-style Kafka messages.
-
-        Parameters
-        ----------
-        raw : `bytes`
-            The 5-byte encoded message prefix.
-
-        Returns
-        -------
-        schema_version : `int`
-            A version number which indicates the Confluent Schema Registry ID
-            number of the Avro schema used to encode the message that follows this
-            header.
-        """
-        ConfluentWireFormatHeader = struct.Struct(">bi")
-        _, version = ConfluentWireFormatHeader.unpack(raw)
-        return version
-
     def _delivery_callback(self, err, msg):
         if err:
-            _log.debug('%% Message failed delivery: %s\n' % err)
+            self.log.warning('Message failed delivery: %s\n' % err)
         else:
-            _log.debug('%% Message delivered to %s [%d] @ %d\n' %
-                       (msg.topic(), msg.partition(), msg.offset()))
+            self.log.debug('Message delivered to %s [%d] @ %d', msg.topic(), msg.partition(), msg.offset())
--- a/python/lsst/ap/association/packageAlerts.py
+++ b/python/lsst/ap/association/packageAlerts.py
@@ -23,11 +23,14 @@ __all__ = ("PackageAlertsConfig", "PackageAlertsTask")
 
 import io
 import os
+import sys
+import logging
 
 from astropy import wcs
 import astropy.units as u
 from astropy.nddata import CCDData, VarianceUncertainty
 import pandas as pd
+import struct
 
 import lsst.alert.packet as alertPack
 import lsst.afw.geom as afwGeom
@@ -36,10 +39,13 @@ import lsst.pex.config as pexConfig
 from lsst.pex.exceptions import InvalidParameterError
 import lsst.pipe.base as pipeBase
 from lsst.utils.timer import timeMethod
-
+import fastavro
 
 """Methods for packaging Apdb and Pipelines data into Avro alerts.
 """
+
+_log = logging.getLogger("lsst." + __name__)
+_log.setLevel(logging.DEBUG)
 
 
 class PackageAlertsConfig(pexConfig.Config):
@@ -63,6 +69,18 @@ class PackageAlertsConfig(pexConfig.Config):
         default=os.path.join(os.getcwd(), "alerts"),
     )
 
+    doProduceAlerts = pexConfig.Field(
+        dtype=bool,
+        doc="Turn on alert production to kafka if true. Set to false by default",
+        default=False,
+    )
+
+    doWriteAlerts = pexConfig.Field(
+        dtype=bool,
+        doc="Write alerts to disk if true. Set to true by default",
+        default=True,
+    )
+
 
 class PackageAlertsTask(pipeBase.Task):
     """Tasks for packaging Dia and Pipelines data into Avro alert packages.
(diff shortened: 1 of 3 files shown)


python/lsst/ap/association/packageAlerts.py
```diff
@@ -36,10 +39,21 @@
...
 import lsst.pipe.base as pipeBase
 from lsst.utils.timer import timeMethod
-
+import fastavro
+from lsst.alert.packet import Schema
+try:
+    from confluent_kafka import KafkaException
+    import confluent_kafka
+except ImportError as error:
+    error.msg += ("Could not import confluent_kafka. Alerts will not be sent to the alert stream")
```
ebellm, line 48: This message isn't logged anywhere...

python/lsst/ap/association/packageAlerts.py
```diff
@@ -36,10 +39,21 @@
...
 
 """Methods for packaging Apdb and Pipelines data into Avro alerts.
 """
+_ConfluentWireFormatHeader = struct.Struct(">bi")
+latest_schema = Schema.from_file().definition
```
ebellm, line 53: Don't load this here--you want to use the `schemaFile` set up in `PackageAlertsConfig`.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -193,6 +244,34 @@ def createDiaSourceExtent(self, bboxSize):
...
+            raise ValueError("Kafka username environment variable was not set.")
+        if not self.server:
+            raise ValueError("Kafka server environment variable was not set.")
+        if not self.kafka_topic:
+            raise ValueError("Kafka topic environment variable was not set.")
+        p = confluent_kafka.Producer(**self.kafka_config)
+        topic = self.kafka_topic
+
+        for alert in alerts:
+            alert_bytes = self._serialize_alert(alert, schema=latest_schema, schema_id=1)
```
ebellm, line 261:
```suggestion
            alert_bytes = self._serialize_alert(alert, schema=self.alertSchema.definition, schema_id=1)
```

python/lsst/ap/association/packageAlerts.py
```diff
@@ -372,3 +451,76 @@ def streamCcdDataToBytes(self, cutout):
             cutout.write(streamer, format="fits")
             cutoutBytes = streamer.getvalue()
         return cutoutBytes
+
+    def _serialize_alert(self, alert, schema=latest_schema, schema_id=0):
```
ebellm, line 455:
```suggestion
    def _serialize_alert(self, alert, schema=self.alertSchema.definition, schema_id=0):
```

python/lsst/ap/association/packageAlerts.py
```diff
@@ -63,6 +77,12 @@ class PackageAlertsConfig(pexConfig.Config):
         default=os.path.join(os.getcwd(), "alerts"),
     )
 
+    produceAlertsConfig = pexConfig.Field(
```
ebellm, line 80:
```suggestion
    doProduceAlerts = pexConfig.Field(
```

python/lsst/ap/association/packageAlerts.py
```diff
@@ -157,10 +203,15 @@ def run(self,
...
                                    templateCutout))
-        with open(os.path.join(self.config.alertWriteLocation,
-                               f"{ccdVisitId}.avro"),
-                  "wb") as f:
-            self.alertSchema.store_alerts(f, alerts)
+
+        if self.config.produceAlertsConfig and "confluent_kafka" in sys.modules:
+            self.produceAlerts(alerts, ccdVisitId)
+
+        else:
```
ebellm, line 210: I'd prefer that we use separate configs for producing and writing alerts: `doProduceAlerts` and `doWriteAlerts`.  Then we can raise an error if `doProduceAlerts` is set and confluent_kafka isn't loaded.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -372,3 +451,76 @@ def streamCcdDataToBytes(self, cutout):
...
+            invalid ID) is used, indicating that the schema is not registered.
+    `
+        Returns
+        -------
+        serialized : `bytes`
+            The byte sequence describing the alert, including the Confluent Wire
+            Format prefix.
+        """
+        buf = io.BytesIO()
+        # TODO: Use a proper schema versioning system
```
ebellm, line 476:
```suggestion
        # TODO: Use a proper schema versioning system (DM-42606)
```

tests/test_packageAlerts.py
```diff
@@ -26,9 +26,15 @@
...
 import unittest
+from unittest.mock import patch, Mock
+import sys
 
 from astropy import wcs
 from astropy.nddata import CCDData
+try:
+    import confluent_kafka
+except ImportError as error:
+    error.msg += ("Could not import confluent_kafka. Alerts will not be sent to the alert stream")
```
ebellm, line 37: Again, this error message isn't being used.

tests/test_packageAlerts.py
ebellm, line 1: We should test applying and interpreting the Confluent wire format header.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -36,10 +39,21 @@
...
+from lsst.alert.packet import Schema
+try:
+    from confluent_kafka import KafkaException
+    import confluent_kafka
+except ImportError as error:
+    error.msg += ("Could not import confluent_kafka. Alerts will not be sent to the alert stream")
 
 """Methods for packaging Apdb and Pipelines data into Avro alerts.
 """
+_ConfluentWireFormatHeader = struct.Struct(">bi")
```
ebellm, line 52: Does this really need to be a global?  Can it live in PackageAlertsTask?

ebellm (changes requested):
looks good overall, a range of suggestions throughout.

python/lsst/ap/association/packageAlerts.py
timj: I think it might be more appropriate to use `error.add_note("Could not import...")` rather than trying to modify the original error message.

python/lsst/ap/association/packageAlerts.py
bsmartradio: Done

python/lsst/ap/association/packageAlerts.py
bsmartradio: Moved this so I could add logging.

python/lsst/ap/association/packageAlerts.py
bsmartradio: Moved `_ConfluentWireFormatHeader` and swapped to `schemaFile`

python/lsst/ap/association/packageAlerts.py
bsmartradio: Done

python/lsst/ap/association/packageAlerts.py
bsmartradio: Done

python/lsst/ap/association/packageAlerts.py
bsmartradio: Swapped to this and added `doWriteAlerts`

python/lsst/ap/association/packageAlerts.py
bsmartradio: This now lives in the PackageAlertTask `_serialize_confluent_wire_header` and `_deserialize` since it isn't needed elsewhere.

tests/test_packageAlerts.py
bsmartradio: Test added as `test_serialize_alert_round_trip`. As per our discussion, there is some concern on the loss of float accuracy during the round trip, as it can vary between 5 and 7 decimal points.

tests/test_packageAlerts.py
bsmartradio: The import has been moved inside the `__init__` to take advantage of `self.logger`.

python/lsst/ap/association/packageAlerts.py
bsmartradio: As mentioned above, we now have both, as well as a check if neither is called set since that would be a situation where nothing happens to the alerts.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -372,3 +465,109 @@ def streamCcdDataToBytes(self, cutout):
...
+        if schema is None:
+            schema = self.alertSchema.definition
+
+        buf = io.BytesIO()
+        # TODO: Use a proper schema versioning system (DM-42606)
+        buf.write(self._serialize_confluent_wire_header(schema_id))
+        fastavro.schemaless_writer(buf, schema, alert)
+        return buf.getvalue()
+
+    def _deserialize_alert(self, alert_bytes, schema=None):
```
parejkoj, line 498: I don't think we want or need this to live in this Task. The purpose of the task is to serialize alerts, not deserialize them. If you need deserialization for tests, you should do that there.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -372,3 +465,109 @@ def streamCcdDataToBytes(self, cutout):
...
+        Notes
+        -----
+        The Confluent Wire Format is described more fully here:
+        https://docs.confluent.io/current/schema-registry/serdes-develop/index.html#wire-format
+        """
+        ConfluentWireFormatHeader = struct.Struct(">bi")
+        return ConfluentWireFormatHeader.pack(0, schema_version)
+
+    @staticmethod
+    def _deserialize_confluent_wire_header(raw):
```
parejkoj, line 549: Same as my comment above: I don't think we want deserialization code in this Task.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -372,3 +465,109 @@ def streamCcdDataToBytes(self, cutout):
...
+            number of the Avro schema used to encode the message that follows this
+            header.
+        """
+        ConfluentWireFormatHeader = struct.Struct(">bi")
+        _, version = ConfluentWireFormatHeader.unpack(raw)
+        return version
+
+    def _delivery_callback(self, err, msg):
+        if err:
+            _log.debug('%% Message failed delivery: %s\n' % err)
```
parejkoj, line 570: If it failed, shouldn't it be `log.error` here or `log.warning`?

tests/test_packageAlerts.py
```diff
@@ -41,6 +43,15 @@
...
 import utils_tests
 
+_log = logging.getLogger("lsst." + __name__)
+_log.setLevel(logging.DEBUG)
+
+try:
+    import confluent_kafka  # noqa: F401
+    from confluent_kafka import KafkaException
+except ModuleNotFoundError as e:
+    _log.error('Kafka module not found: {}'.format(e))
```
parejkoj, line 53: How many of the tests here require kafka? You'll want to use a `skipIf` decorator on the test methods that require kafka, so that they are explicitly skipped. Having a logger here isn't necessary, if you decorate the tests and skip them with an appropriate message.

tests/test_packageAlerts.py
```diff
@@ -95,9 +110,46 @@ def _roundTripThroughApdb(objects, sources, forcedSources, dateTime):
     return (diaObjects, diaSources, diaForcedSources)
 
 
+def mock_alert(alert_id):
+    """Generate a minimal mock alert.
+    """
+    return {
+        "alertId": alert_id,
+        "diaSource": {
+            # Below are all the required fields containing random values.
```
parejkoj, line 119: What purpose does this comment serve?

tests/test_packageAlerts.py
```diff
@@ -95,9 +110,46 @@ def _roundTripThroughApdb(objects, sources, forcedSources, dateTime):
...
+    }
+
+
 class TestPackageAlerts(lsst.utils.tests.TestCase):
+    kafka_enabled = "confluent_kafka" in sys.modules
+
+    def __init__(self, *args, **kwargs):
+        TestPackageAlerts.kafka_enabled = "confluent_kafka" in sys.modules
+        _log.debug('TestPackageAlerts: kafka_enabled={}'.format(self.kafka_enabled))
+        super(TestPackageAlerts, self).__init__(*args, **kwargs)
```
parejkoj, line 144: I don't think you want this at all: you want to set a global in the `except ModuleNotFoundError as e:` above and decorate either the whole class (if all tests in the class need kafka) or the kafka-specific methods with `@unittest.skipIf`.

https://docs.python.org/3/library/unittest.html#unittest.skipIf

tests/test_packageAlerts.py
```diff
@@ -275,9 +330,122 @@ def testMakeAlertDict(self):
...
+            packConfig = PackageAlertsConfig(doProduceAlerts=True)
+            packageAlerts = PackageAlertsTask(config=packConfig)  # noqa: F841
+
+    def test_produceAlerts_empty_username(self):
+        """ Test that produceAlerts raises if the username is empty or missing.
+        """
+        self.environ['AP_KAFKA_PRODUCER_USERNAME'] = ""
+        with self.assertRaisesRegex(ValueError, "Kafka username"):
+            packConfig = PackageAlertsConfig(doProduceAlerts=True)
+            packageAlerts = PackageAlertsTask(config=packConfig)  # noqa: F841
```
parejkoj, line 352: Don't assign to `packageAlerts` here: the raised exception means that the task instantiation doesn't return anything. Then you don't have to have noqa for all of these.

parejkoj (changes requested):
I've made a lot of cleanup comments. I didn't look at all of the test: make the `skipIf` decorator cleanups and I'll look at the tests more.

You have a mix of snake_case and camelCase in the changes here: please be consistent in this package. The existing code is camelCase, so please use that for all variables and functions.

parejkoj (commented):
How do we run the tests that depend on kafka, if kafka isn't in `rubin-env`? Is it in `rubin-env-developer`?

bsmartradio:
It is in my rubin-dvl, I cannot remember if I set that up or if its present for everyone.
________________________________

python/lsst/ap/association/packageAlerts.py
```diff
@@ -77,6 +95,53 @@ def __init__(self, **kwargs):
...
+                "sasl.password": self.password,
+                # Batch size limits the largest size of a kafka alert that can be sent.
+                # We set the batch size to 2 Mb.
+                "batch.size": 2097152,
+                "linger.ms": 5,
+            }
+
+            try:
+                from confluent_kafka import KafkaException
+                self.kafka_exception = KafkaException
```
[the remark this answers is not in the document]
bsmartradio: KafkaException doesn't seem to be visible where I need it when it is imported here and not with the initial imports. If there is a workaround for this, I'm happy to use that instead.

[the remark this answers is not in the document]
bsmartradio: Changed to Eli's recommendation, import now at the top and `confluent_kafka` is set to `None` if it can't import. Within the init, it now starts by checking `if confluent_kafka is not None` it then sets up the alert production, otherwise it skips it all.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -77,6 +95,53 @@ def __init__(self, **kwargs):
...
+                _log.error(error)
+
+            if not self.password:
+                raise ValueError("Kafka password environment variable was not set.")
+            if not self.username:
+                raise ValueError("Kafka username environment variable was not set.")
+            if not self.server:
+                raise ValueError("Kafka server environment variable was not set.")
+            if not self.kafka_topic:
+                raise ValueError("Kafka topic environment variable was not set.")
```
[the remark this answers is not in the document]
bsmartradio: I tried your suggested code, it does not like the self.password within the walrus operator. Googling, walrus operators aren't allowed with instance attributes. https://stackoverflow.com/questions/64055314/why-cant-pythons-walrus-operator-be-used-to-set-instance-attributes

python/lsst/ap/association/packageAlerts.py
```diff
@@ -157,10 +222,21 @@ def run(self,
                                    objDiaForcedSources,
                                    diffImCutout,
                                    templateCutout))
-        with open(os.path.join(self.config.alertWriteLocation,
-                               f"{ccdVisitId}.avro"),
-                  "wb") as f:
-            self.alertSchema.store_alerts(f, alerts)
+
+        if self.config.doProduceAlerts and "confluent_kafka" in sys.modules:
```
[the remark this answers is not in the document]
bsmartradio: Swapped to setting `self.config.doProduceAlerts` first and then it checks `if confluent_kafka is not None:`, if not, it logs an error message.

tests/test_packageAlerts.py
bsmartradio: Removed

tests/test_packageAlerts.py
bsmartradio: All skipIfs now check if `confluent_kafka = None` which now happens in the imports.

tests/test_packageAlerts.py
bsmartradio: Removed and swapped to Eli's method of setting confluent_kafka to None.

tests/test_packageAlerts.py
bsmartradio: Fixed

tests/test_packageAlerts.py
bsmartradio: The import now is back with the main imports and sets `confluent_kafka = None` when it fails. This then triggers an error message to be logged if `doProduceAlerts` was set.

tests/test_packageAlerts.py
bsmartradio: Test is now fixed with setting variables explicitly to 32 bit floats.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -63,6 +69,18 @@ class PackageAlertsConfig(pexConfig.Config):
...
 
+    doProduceAlerts = pexConfig.Field(
+        dtype=bool,
+        doc="Turn on alert production to kafka if true. Set to false by default",
+        default=False,
+    )
+
+    doWriteAlerts = pexConfig.Field(
+        dtype=bool,
+        doc="Write alerts to disk if true. Set to true by default",
```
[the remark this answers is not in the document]
bsmartradio: They are both independent of one another, and can both be set or one or the other. An error only needs to be set if both are not set.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -63,6 +67,18 @@ class PackageAlertsConfig(pexConfig.Config):
...
+    doProduceAlerts = pexConfig.Field(
+        dtype=bool,
+        doc="Turn on alert production to kafka if true and if confluent_kafka is in the environment.",
+        default=False,
+    )
+
+    doWriteAlerts = pexConfig.Field(
+        dtype=bool,
+        doc="Write alerts to disk if true.",
+        default=True,
```
ebellm, line 79:
```suggestion
        default=False,
```

This should default to false to match the existing behavior.

Also, I just appreciated that we need to formally deprecate `doPackageAlerts` rather than simply removing it: see https://developer.lsst.io/stack/deprecating-interfaces.html#config-deprecation

also needs an RFC: https://developer.lsst.io/communications/rfc.html#standard-procedures-that-require-an-rfc

python/lsst/ap/association/packageAlerts.py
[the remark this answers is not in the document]
ebellm: I agree that the proposed `is not None` is better but I don't see it in the code?

tests/test_packageAlerts.py
```diff
@@ -153,6 +204,49 @@ def setUp(self):
...
+        -------
+        alert : `dict`
+            An alert payload.
+        """
+        schema = alertPack.Schema.from_uri(
+            str(alertPack.get_uri_to_latest_schema()))
+
+        header_bytes = alert_bytes[:5]
+        version = self._deserialize_confluent_wire_header(header_bytes)
+        assert version == 0
```
ebellm, line 226: `self.assertEqual` instead?

python/lsst/ap/association/packageAlerts.py
ebellm: If we wanted to avoid the song and dance we could just continue to use `doPackageAlerts` in place of `doWriteAlerts`.  It's less clear but less hassle.

ebellm (changes requested):
Getting there!  A couple more small items.

python/lsst/ap/association/packageAlerts.py
bsmartradio: I'll go this way for now.

tests/test_packageAlerts.py
bsmartradio: Missed changing that one, thanks for pointing it out. Swapped to `self.assertEqual`

python/lsst/ap/association/packageAlerts.py
```diff
@@ -77,6 +93,52 @@ def __init__(self, **kwargs):
...
+                    # We set the batch size to 2 Mb.
+                    "batch.size": 2097152,
+                    "linger.ms": 5,
+                }
+
+                self.producer = confluent_kafka.Producer(**self.kafkaConfig)
+
+            else:
+                self.log.error("Produce alerts is set but confluent_kafka is not present in "
+                               "the environment. Alerts will not be sent to the alert stream.")
```
parejkoj, line 140: Shouldn't this be a `raise` (probably RuntimeError for now?)? If someone sets that to True, they want to produce alerts.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -77,6 +93,52 @@ def __init__(self, **kwargs):
         self.alertSchema = alertPack.Schema.from_uri(self.config.schemaFile)
         os.makedirs(self.config.alertWriteLocation, exist_ok=True)
 
+        if self.config.doProduceAlerts:
+
+            if confluent_kafka is not None:
+
```
parejkoj, line 99: Unnecessary line.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -77,6 +93,52 @@ def __init__(self, **kwargs):
...
+                self.kafkaTopic = os.getenv("AP_KAFKA_TOPIC")
+
+                if not self.password:
+                    raise ValueError("Kafka password environment variable was not set.")
+                if not self.username:
+                    raise ValueError("Kafka username environment variable was not set.")
+                if not self.server:
+                    raise ValueError("Kafka server environment variable was not set.")
+                if not self.kafkaTopic:
+                    raise ValueError("Kafka topic environment variable was not set.")
```
parejkoj, line 112: Since you can't do these with a walrus, at least put each `if` directly under the value that is being assigned to.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -63,6 +67,18 @@ class PackageAlertsConfig(pexConfig.Config):
         default=os.path.join(os.getcwd(), "alerts"),
     )
 
+    doProduceAlerts = pexConfig.Field(
+        dtype=bool,
+        doc="Turn on alert production to kafka if true and if confluent_kafka is in the environment.",
+        default=False,
+    )
+
+    doPackageAlerts = pexConfig.Field(
```
parejkoj, line 76: Should this be `doWriteAlerts` instead? "doProduce" and "doPackage" sound very similar.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -157,10 +223,13 @@ def run(self,
                                    objDiaForcedSources,
                                    diffImCutout,
                                    templateCutout))
-        with open(os.path.join(self.config.alertWriteLocation,
-                               f"{ccdVisitId}.avro"),
-                  "wb") as f:
-            self.alertSchema.store_alerts(f, alerts)
+
+        if self.config.doProduceAlerts and confluent_kafka is not None:
```
parejkoj, line 227: If we `raise` in `__init__` for this case like I suggest above, the latter part of this is unnecessary.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -173,7 +242,7 @@ def _patchDiaSources(self, diaSources):
         diaSources["programId"] = 0
 
     def createDiaSourceExtent(self, bboxSize):
-        """Create a extent for a  box for the cutouts given the size of the
+        """Create an extent for a  box for the cutouts given the size of the
```
parejkoj, line 245:
```suggestion
        """Create an extent for a box for the cutouts given the size of the
```

python/lsst/ap/association/packageAlerts.py
```diff
@@ -63,6 +67,18 @@ class PackageAlertsConfig(pexConfig.Config):
         default=os.path.join(os.getcwd(), "alerts"),
```
parejkoj, line 67: Oh dear, I just noticed this: we definitely don't want it to default to writing to the current working directory, especially since we're using it as the dumping ground for alerts that fail to be sent via kafka! I think this path should either have no default, or be a tempdir.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -372,3 +458,63 @@ def streamCcdDataToBytes(self, cutout):
...
+        -----
+        The Confluent Wire Format is described more fully here:
+        https://docs.confluent.io/current/schema-registry/serdes-develop/index.html#wire-format
+        """
+        ConfluentWireFormatHeader = struct.Struct(">bi")
+        return ConfluentWireFormatHeader.pack(0, schema_version)
+
+    def _delivery_callback(self, err, msg):
+        if err:
+            self.log.debug('%% Message failed delivery: %s\n' % err)
```
parejkoj, line 517: I think errors should be logged as at least info, if not warning level? Also, use [lazy formatting](https://developer.lsst.io/stack/logging.html#basic-usage-in-python):

```
self.log.debug('%% Message failed delivery: %s', err)
```

parejkoj: Why the prefaced `%` in these two logs?

python/lsst/ap/association/packageAlerts.py
```diff
@@ -372,3 +458,63 @@ def streamCcdDataToBytes(self, cutout):
...
+        else:
+            self.log.debug('%% Message delivered to %s [%d] @ %d\n'
+                           % (msg.topic(), msg.partition(), msg.offset()))
```
parejkoj, line 520: Use [lazy formatting](https://developer.lsst.io/stack/logging.html#basic-usage-in-python):

```suggestion
            self.log.debug('%% Message delivered to %s [%d] @ %d', msg.topic(), msg.partition(), msg.offset())
```

tests/test_packageAlerts.py
```diff
@@ -95,9 +108,44 @@ def _roundTripThroughApdb(objects, sources, forcedSources, dateTime):
...
+            "flags": 12345,
+        }
+    }
+
+
 class TestPackageAlerts(lsst.utils.tests.TestCase):
 
+    def __init__(self, *args, **kwargs):
+        super(TestPackageAlerts, self).__init__(*args, **kwargs)
+
```
parejkoj, line 141: Unnecessary, please remove.

tests/test_packageAlerts.py
```diff
@@ -153,6 +204,49 @@ def setUp(self):
...
+        -------
+        alert : `dict`
+            An alert payload.
+        """
+        schema = alertPack.Schema.from_uri(
+            str(alertPack.get_uri_to_latest_schema()))
+
+        header_bytes = alert_bytes[:5]
+        version = self._deserialize_confluent_wire_header(header_bytes)
+        self.assertEqual(version, 0)
```
parejkoj, line 226: What is the purpose of this assert?

tests/test_packageAlerts.py
```diff
@@ -153,6 +204,49 @@ def setUp(self):
         self.cutoutWcs.wcs.cd = self.exposure.getWcs().getCdMatrix()
         self.cutoutWcs.wcs.ctype = ["RA---TAN", "DEC--TAN"]
 
+    def _deserialize_alert(self, alert_bytes):
```
parejkoj, line 207: I think this, and the `_deserialize_confluent_wire_header` should both be free functions here: they don't need to live in the test class at all.

tests/test_packageAlerts.py
parejkoj: I think the purpose of that bare `assert` was to check that the version coming out of the header is always 0. I don't know if that's actually the behavior we want here, but I don't think it's a unittest-related thing, but rather a "is the code sensible" test, hence a bare assert. I might be wrong, though?

parejkoj (changes requested):
More changes.

parejkoj:
There are some comments from the previous review that were apparently missed: I marked them with 👀 .

python/lsst/ap/association/packageAlerts.py
```diff
@@ -23,11 +23,19 @@
...
 
 from astropy import wcs
 import astropy.units as u
 from astropy.nddata import CCDData, VarianceUncertainty
 import pandas as pd
+import struct
+import fastavro
+try:
+    import confluent_kafka
+    from confluent_kafka import KafkaException
```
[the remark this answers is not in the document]
bsmartradio: Added ```# confluent_kafka is not in the standard Rubin environment as it is a third
# party package and is only needed when producing alerts.```

python/lsst/ap/association/packageAlerts.py
bsmartradio: Done

python/lsst/ap/association/packageAlerts.py
bsmartradio: I assume then it would be fine to remove the `self.log.error` since `raise` would be logged?

python/lsst/ap/association/packageAlerts.py
```diff
@@ -193,6 +269,23 @@ def createDiaSourceExtent(self, bboxSize):
             extent = geom.Extent2I(bboxSize, bboxSize)
         return extent
 
+    def produceAlerts(self, alerts, ccdVisitId):
+
```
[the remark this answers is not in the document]
bsmartradio: Added docstrings.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -372,3 +465,109 @@ def streamCcdDataToBytes(self, cutout):
...
+        ----------`
+        alert : `dict`
+            An alert payload to be serialized.
+        schema : `dict`, optional
+            An Avro schema definition describing how to encode `alert`. By default,
+            the schema is None, which sets it to the latest schema available.
+        schema_id : `int`, optional
+            The Confluent Schema Registry ID of the schema. By default, 0 (an
+            invalid ID) is used, indicating that the schema is not registered.
+    `
```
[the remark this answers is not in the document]
bsmartradio: I'm not sure what you mean by ?? Do you mean I should do the serialization within produceAlerts?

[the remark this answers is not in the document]
parejkoj: There's an errant ` on this line.

python/lsst/ap/association/packageAlerts.py
parejkoj: Yes, if we just raise we don't need to log.

tests/test_packageAlerts.py
bsmartradio: Its a sanity check leftover to make sure the version was correct. I can remove it since it isn't necessary for the unit checks.

tests/test_packageAlerts.py
bsmartradio: I've removed `_deserialized_confluent_wire_header` because its not needed if we aren't doing the header version sanity check.

python/lsst/ap/association/packageAlerts.py
bsmartradio: Leftover formatting from when this lived in a different package. It's been removed.

python/lsst/ap/association/packageAlerts.py
bsmartradio: Looking at the package again, I think we can actually leave `doWriteAlerts` as `doPackageAlerts` is actually at the diaPipe level and not in the `PackageAlertsTask`? Do `doPackageAlerts` gets set in dia pipe, then inside there `doWriteAlerts` or `doProduceAlerts` gets set. There was never a config to turn on or off writing the alerts out.

ebellm (approved):

bsmartradio:
Jenkins run: https://rubin-ci.slac.stanford.edu/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/774/pipeline

python/lsst/ap/association/packageAlerts.py
```diff
@@ -193,6 +260,33 @@ def createDiaSourceExtent(self, bboxSize):
             extent = geom.Extent2I(bboxSize, bboxSize)
         return extent
 
+    def produceAlerts(self, alerts, ccdVisitId):
+        """Serialize alerts and send them to the alert stream using
+        confluent_kafka's producer.
+
+                Parameters
+        ----------
```
parejkoj, line 268: These aren't lined up.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -372,3 +466,62 @@ def streamCcdDataToBytes(self, cutout):
             cutout.write(streamer, format="fits")
             cutoutBytes = streamer.getvalue()
         return cutoutBytes
+
+    def _serializeAlert(self, alert, schema=None, schema_id=0):
+        """Serialize an alert to a byte sequence for sending to Kafka.
+
+        Parameters
+        ----------`
```
parejkoj, line 474: Excess ` at the end

parejkoj (approved):
A couple remaining comments about docstrings, but otherwise this looks much better, thank you.

python/lsst/ap/association/packageAlerts.py
timj: If people are interested, I have been experimenting with numpydoc validation in middleware packages and can point people at that. Also, the velin package can fix some problems automatically.

bsmartradio: @timj Yes I definitely would be interested in that!

timj: `velin` fixes automatically both of the problems that @parejkoj points out above. It also fixes some other indent issues and variable name and section headings. We don't install `velin` in the shared stack by default (it's not in rubin-develop-env but maybe it could be cc/ @ktlim ).

`numpydoc` is also not in the rubin env but has many more extensive checks. It also needs to have a lot of configuration. The quickest way to see what you need to do is to look at lsst/daf_butler#922 and see the changes I made to `pyproject.toml` and the github workflows.

You can run the validation with:

```bash
$ python -m numpydoc.hooks.validate_docstrings $(find python -name "*.py")
```
after you've installed `numpydoc`. This will definitely need some tuning in your config files to keep the checks you think are useful.

PR #197, opened 2024-03-20
DM-43389: Add option to disable loading diaForcedSource history

isullivan:
This is a debugging-only option. I elected to pass `doLoadForcedSources` in to `LoadDiaCatalogsTask.run()` rather than set it as a config of the subtask, thinking that it would be easier to remove in the future this way. I would welcome alternate suggestions!

### DIFF
--- a/python/lsst/ap/association/diaPipe.py
+++ b/python/lsst/ap/association/diaPipe.py
@@ -244,6 +244,15 @@ class DiaPipelineConfig(pipeBase.PipelineTaskConfig,
         target=DiaObjectCalculationTask,
         doc="Task to compute summary statistics for DiaObjects.",
     )
+    doLoadForcedSources = pexConfig.Field(
+        dtype=bool,
+        default=True,
+        deprecated="Added to allow disabling forced sources for performance"
+                   "reasons during the ops rehearsal."
+                   " It is expected to be removed.",
+        doc="Load forced DiaSource history from the APDB?"
+            "This should only be turned off for debugging purposes.",
+    )
     diaForcedSource = pexConfig.ConfigurableField(
         target=DiaForcedSourceTask,
         doc="Task used for force photometer DiaObject locations in direct and "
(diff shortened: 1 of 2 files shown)


python/lsst/ap/association/diaPipe.py
```diff
@@ -244,6 +244,12 @@ class DiaPipelineConfig(pipeBase.PipelineTaskConfig,
         target=DiaObjectCalculationTask,
         doc="Task to compute summary statistics for DiaObjects.",
     )
+    doLoadForcedSources = pexConfig.Field(
+        dtype=bool,
+        default=True,
+        doc="Load forced DiaSource history from the APDB?"
+            "This should only be turned off for debugging purposes.",
+    )
```
parejkoj, line 252: I feel like this docstring should say something about how this was added to let us turn off forced sources for performance reasons? Maybe something like this?

```suggestion
        doc="Load forced DiaSource history from the APDB?"
            " This should only be turned off for debugging purposes."
            " Added to allow disabling forced sources for performance reasons during the ops rehearsal.",
    )
```

python/lsst/ap/association/loadDiaCatalogs.py
```diff
@@ -66,6 +66,9 @@ def run(self, exposure, apdb):
             An exposure with a bounding box.
         apdb : `lsst.dax.apdb.Apdb`
             AP database connection object.
+        doLoadForcedSources : `bool`, optional
+            Load forced DiaSource history from the APDB?
+            This should only be turned off for debugging purposes.
```
parejkoj, line 71: Same note as above with the config docstring.

parejkoj (commented):
Couple of comments about docstrings, but otherwise I think this is ok. I do wish there was a way to test it in this code, instead of relying on ap_verify, but I'm not sure there's a good way to do that.

python/lsst/ap/association/loadDiaCatalogs.py
```diff
@@ -79,6 +82,11 @@ def run(self, exposure, apdb):
               exposure padded by ``pixelMargin``. DataFrame is indexed by
               ``diaObjectId``, ``band``, ``diaSourceId`` columns.
               (`pandas.DataFrame`)
+
```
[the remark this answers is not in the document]
isullivan: Ha. That's been missing for a while... I'll add it.

PR #198, opened 2024-03-20
DM-43402: Add option to turn off forced measurements on the science and differe…

isullivan:
…nce images

### DIFF
--- a/python/lsst/ap/association/diaPipe.py
+++ b/python/lsst/ap/association/diaPipe.py
@@ -130,6 +130,8 @@ class DiaPipelineConnections(
             self.outputs.remove("associatedDiaSources")
             self.outputs.remove("diaForcedSources")
             self.outputs.remove("diaObjects")
+        elif not config.doRunForcedMeasurement:
+            self.outputs.remove("diaForcedSources")
         if not config.doSolarSystemAssociation:
             self.inputs.remove("solarSystemObjectTable")
         if not config.associator.doTrailedSourceFilter:
@@ -247,10 +249,19 @@ class DiaPipelineConfig(pipeBase.PipelineTaskConfig,
     doLoadForcedSources = pexConfig.Field(
         dtype=bool,
         default=True,
-        deprecated="Added to allow disabling forced sources for performance"
-                   "reasons during the ops rehearsal."
-                   " It is expected to be removed.",
-        doc="Load forced DiaSource history from the APDB?"
+        deprecated="Added to allow disabling forced sources for performance "
+                   "reasons during the ops rehearsal. "
+                   "It is expected to be removed.",
+        doc="Load forced DiaSource history from the APDB? "
+            "This should only be turned off for debugging purposes.",
+    )
+    doRunForcedMeasurement = pexConfig.Field(
+        dtype=bool,
+        default=True,
+        deprecated="Added to allow disabling forced sources for performance "
+                   "reasons during the ops rehearsal. "
+                   "It is expected to be removed.",
+        doc="Run forced measurement on all of the diaObjects? "
             "This should only be turned off for debugging purposes.",
     )
     diaForcedSource = pexConfig.ConfigurableField(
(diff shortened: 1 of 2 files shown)


python/lsst/ap/association/diaPipe.py
```diff
@@ -498,14 +507,17 @@ def run(self,
...
+            # Force photometer on the Difference and Calibrated exposures using
+            # the new and updated DiaObject locations.
+            diaForcedSources = self.diaForcedSource.run(
+                diaCalResult.diaObjectCat,
+                diaCalResult.updatedDiaObjects.loc[:, "diaObjectId"].to_numpy(),
+                exposure,
+                diffIm,
+                idGenerator=idGenerator)
+        else:
+            diaForcedSources = pd.DataFrame(columns=["diaObjectId", "diaForcedSourceId"])
```
kfindeisen, line 520: I don't think this is correct; surely `DiaForcedSourceTask` has columns for the fluxes (at minimum, [these ones](https://github.com/lsst/ap_association/blob/main/python/lsst/ap/association/diaForcedSource.py#L219-L223))? Unfortunately, `DiaForcedSourceTask` doesn't have a mechanism for reporting its output schema, so fixing this may be tricky.

Pinging @parejkoj for a second opinion, since he knows catalog datasets much better than I do.

python/lsst/ap/association/diaPipe.py
```diff
@@ -557,7 +569,9 @@ def run(self,
                                    diaForcedSources,
                                    diffIm,
                                    exposure,
-                                   template)
+                                   template,
+                                   doUseForcedMeasurements=self.config.doRunForcedMeasurement,
```
kfindeisen, line 573:
```suggestion
                                   doRunForcedMeasurement=self.config.doRunForcedMeasurement,
```

kfindeisen (changes requested):
The concept seems sound enough, but I don't think this code was actually tested, and will break (either when calling `Apdb.store` or `alertPackager.run`) if we try to run it...

python/lsst/ap/association/diaPipe.py
kfindeisen: Confirmed that it doesn't break `Apdb.store`, but the dummy catalog gets written to disk. Will fix.

parejkoj: Unfortunately, this is a pandas question, and I don't know what our schema requirements are for our pandas outputs, or whether such a thing is even define-able, given pandas' lack of schema. I'd assume it needs at minimum the indexes defined (which I don't think this does?).

kfindeisen: Reading the dummy catalog from Butler restores the columns but no index.

kfindeisen:
With the obvious issues fixed, `packageAlerts` chokes on a schema mismatch. Consulting with @bsmartradio...

python/lsst/ap/association/diaPipe.py
```diff
@@ -253,6 +255,15 @@ class DiaPipelineConfig(pipeBase.PipelineTaskConfig,
...
             "This should only be turned off for debugging purposes.",
     )
+    doRunForcedMeasurement = pexConfig.Field(
+        dtype=bool,
+        default=True,
+        deprecated="Added to allow disabling forced sources for performance"
+                   "reasons during the ops rehearsal."
+                   " It is expected to be removed.",
+        doc="Run forced measurement on all of the diaObjects?"
+            "This should only be turned off for debugging purposes.",
```
parejkoj, line 265:
```suggestion
        deprecated="Added to allow disabling forced sources for performance"
                   " reasons during the ops rehearsal."
                   " It is expected to be removed.",
        doc="Run forced measurement on all of the diaObjects?"
            " This should only be turned off for debugging purposes.",
```

parejkoj (approved):
One spacing cleanup, otherwise my same concern as the earlier ticket about this (that we don't have a way to really test it here), but I won't stop that from merging it, since this is just a stopgap.

PR #201, opened 2024-03-26
DM-43514: Fix bug in APDB bit flag unpacking

parejkoj:


python/lsst/ap/association/transformDiaSourceCatalog.py
```diff
@@ -378,12 +378,13 @@ def unpack(self, input_flag_values, flag_name):
...
         """
-        bit_names_types = self.output_flag_columns[flag_name]
-        output_flags = np.zeros(len(input_flag_values), dtype=bit_names_types)
+        output_flags = np.zeros(len(input_flag_values),
+                                dtype=[(name, bool) for name in self.output_flag_columns[flag_name]])
 
-        for bit_idx, (bit_name, dtypes) in enumerate(bit_names_types):
-            masked_bits = np.bitwise_and(input_flag_values, 2**bit_idx)
-            output_flags[bit_name] = masked_bits
+        for name in self.output_flag_columns[flag_name]:
```
mrawls, line 384: You're basically iterating over column names, right? This might be clearer if the iterable were, accordingly, `columnName`, here and in the list comp above, as it is in the `flagExists` method below

Of course, camelCase vs snake_case is rearing its head, maybe it should be `column_name` throughout...

python/lsst/ap/association/transformDiaSourceCatalog.py
```diff
@@ -378,12 +378,13 @@ def unpack(self, input_flag_values, flag_name):
         output_flags : `numpy.ndarray`
             Numpy named tuple of booleans.
         """
-        bit_names_types = self.output_flag_columns[flag_name]
-        output_flags = np.zeros(len(input_flag_values), dtype=bit_names_types)
+        output_flags = np.zeros(len(input_flag_values),
+                                dtype=[(name, bool) for name in self.output_flag_columns[flag_name]])
```
mrawls, line 382: I figured out what this does after falling down a `dtype` rabbit hole, but, possibly silly question, aren't they all required (by virtue of being flags) to just be `bool`? Is there a reason we bend over backwards to assign the dtypes one at a time?

python/lsst/ap/association/transformDiaSourceCatalog.py
```diff
@@ -356,9 +356,9 @@ def __init__(self, flag_map_file, table_name):
         self.output_flag_columns = {}
 
         for column in self.bit_pack_columns:
-            names = []
+            names = {}
             for bit in column["bitList"]:
-                names.append((bit["name"], bool))
+                names[bit["name"]] = bit['bit']
```
mrawls, line 361: Odd single vs double quote mismatch

python/lsst/ap/association/transformDiaSourceCatalog.py
```diff
@@ -378,12 +378,13 @@ def unpack(self, input_flag_values, flag_name):
...
-        output_flags = np.zeros(len(input_flag_values), dtype=bit_names_types)
+        output_flags = np.zeros(len(input_flag_values),
+                                dtype=[(name, bool) for name in self.output_flag_columns[flag_name]])
 
-        for bit_idx, (bit_name, dtypes) in enumerate(bit_names_types):
-            masked_bits = np.bitwise_and(input_flag_values, 2**bit_idx)
-            output_flags[bit_name] = masked_bits
+        for name in self.output_flag_columns[flag_name]:
+            masked_bits = np.bitwise_and(input_flag_values,
+                                         2**self.output_flag_columns[flag_name][name])
```
mrawls, line 386: I think this could be clearer if you defined `bit_idx = self.output_flag_columns[flag_name][name]` first and then set `masked_bits` using `bit_idx`, but I don't insist on it.

mrawls (approved):
As best I can tell this fixes the bug and passes the tests, thank you, all my comments are for clarity and not functionality.

python/lsst/ap/association/transformDiaSourceCatalog.py
parejkoj: I have no idea why Chris originally made this a dict of tuples of `(name, bool)`, because they're all booleans because this is a bit field. That's why I changed that part of it.

I put in a dtype for the numpy array, so that it is a structured array with column names, each one a bool.

python/lsst/ap/association/transformDiaSourceCatalog.py
parejkoj: It's not `bit_idx`, though, it's the bit. It's not an index.

python/lsst/ap/association/transformDiaSourceCatalog.py
parejkoj: They're the name of the thing in the dictionary, so it's just `name`. It's generally better to keep iterator variables short.

And yeah, Chris was unconsistent with camel vs. snakes. 🤷

parejkoj:
I tried to cleanup some of the docstrings to help clarify `unpack`.

PR #202, opened 2024-03-26
DM-43522:Remove too large alert packet write to disk

bsmartradio:
As writing failed alerts to disk is only used for debugging, it should not be on by default and instead a configuration should be set when we want to inspect the alerts.

### DIFF
--- a/python/lsst/ap/association/packageAlerts.py
+++ b/python/lsst/ap/association/packageAlerts.py
@@ -84,8 +84,8 @@ class PackageAlertsConfig(pexConfig.Config):
 
     doWriteFailedAlerts = pexConfig.Field(
         dtype=bool,
-        doc="If an alert cannot be sent when doProduceAlerts is set,"
-            " write it to disk for debugging purposes.",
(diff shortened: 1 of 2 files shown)


python/lsst/ap/association/packageAlerts.py
```diff
@@ -82,6 +82,12 @@ class PackageAlertsConfig(pexConfig.Config):
         default=False,
     )
 
+    doWriteFailedAlerts = pexConfig.Field(
+        dtype=bool,
+        doc="Write alerts which fail to send to disk for debugging purposes.",
```
kfindeisen, line 87: It's probably worth mentioning that this option is only used if `doProduceAlerts`, and is independent of `doWriteAlerts`.

kfindeisen: I find the wording a bit hard to parse; maybe:
```suggestion
        doc="If an alert cannot be sent, write it to disk for debugging purposes.",
```

kfindeisen (approved):
Looks good, just some doc suggestions.

PR #203, opened 2024-03-27
DM-43509: Switch to astropy.time in calls to Apdb methods

andy-slac:
APDB interface replaced the use of daf_base.DateTime with astropy Time, this patch updates client code to convert DateTtime to astropy Time when passing arguments to Apdb.


parejkoj (approved):
Fewer places needed changing than I would have guessed!

PR #204, opened 2024-04-01
DM-43590: Add timeout config field

bsmartradio:
Add timeout config to packageAlerts and set the default to 15 seconds. The current static timeout appears to be too short, causing timeouts when the pipeline initializes and intermittently when alerts are being sent.


bsmartradio:
https://rubin-ci.slac.stanford.edu/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/1051/pipeline

ebellm (approved):

PR #206, opened 2024-04-09
DM-43501: Add delivery timeout config

bsmartradio:
Add a delivery timeout config to package alerts. Removed check_server from before the alerts are sent as it now timeouts very quickly if there are any sending issues.


ebellm (approved):
Looks good.

PR #207, opened 2024-04-19
DM-42744: Fix random seeds for `ap_association` unit tests

BrunoSanchez:
Set up random generator seed in two places, main location is the `utils_tests.py` that is reused broadly.

## Checklist

- [x] ran Jenkins
- [x] ran local unit tests

### DIFF
--- a/tests/utils_tests.py
+++ b/tests/utils_tests.py
@@ -33,7 +33,7 @@ import lsst.daf.base as dafBase
 import lsst.geom
 
 
-def makeDiaObjects(nObjects, exposure):
+def makeDiaObjects(nObjects, exposure, rng):
     """Make a test set of DiaObjects.
 
     Parameters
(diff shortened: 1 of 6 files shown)


tests/utils_tests.py
```diff
@@ -32,6 +32,9 @@
 import lsst.daf.base as dafBase
 import lsst.geom
 
+# Create an instance of random generator with fixed seed.
+rng = np.random.default_rng(1234)
```
isullivan, line 36: In my testing, this results in different random catalogs for each test if I randomize the order the tests are run with `pytest --random-order`. I suggest instead defining `rng` in the `setUp` method of each class that needs it, and passing that in to `makeDiaObjects`, `makeDiaSources`, and `makeDiaForcedSources`

isullivan (changes requested):
See the code comment.

BrunoSanchez:
Changes implemented. Ready for new review

tests/utils_tests.py
```diff
@@ -32,8 +32,11 @@
 import lsst.daf.base as dafBase
 import lsst.geom
 
+# # Create an instance of random generator with fixed seed.
+# rng = np.random.default_rng(1234)
```
isullivan, line 36: Did you intend for this to be commented out?

isullivan (approved):
One commented-out line should probably be uncommented, or removed if not needed.

tests/utils_tests.py
BrunoSanchez: No sorry. Thanks for the catch

PR #208, opened 2024-04-23
DM-43856: Debug running AP with Cassandra in batch mode

isullivan:
This PR is not intended to be merged yet. It is an example of the changes needed to run the AP pipeline with a Cassandra APDB.

### DIFF
--- a/python/lsst/ap/association/transformDiaSourceCatalog.py
+++ b/python/lsst/ap/association/transformDiaSourceCatalog.py
@@ -24,9 +24,10 @@ __all__ = ("TransformDiaSourceCatalogConnections",
            "TransformDiaSourceCatalogTask",
            "UnpackApdbFlags")
 
-import numpy as np
 import os
 import yaml
+
+import numpy as np
 import pandas as pd
 
 from lsst.daf.base import DateTime
@@ -36,6 +37,8 @@ import lsst.pipe.base.connectionTypes as connTypes
 from lsst.pipe.tasks.postprocess import TransformCatalogBaseTask, TransformCatalogBaseConfig
 from lsst.pipe.tasks.functors import Column
 from lsst.utils.timer import timeMethod
+
+from .utils import convertTableToSdmSchema, readSdmSchemaFile
 
 
 class TransformDiaSourceCatalogConnections(pipeBase.PipelineTaskConnections,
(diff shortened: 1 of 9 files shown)


python/lsst/ap/association/utils.py
```diff
@@ -0,0 +1,72 @@
...
+        if columnDef.name in sourceTable.columns:
+            dtype = apdbSchema.column_dtype(columnDef.datatype)
+            data[columnDef.name] = pd.Series(sourceTable[columnDef.name], dtype=dtype)
+        else:
+            dataInit = np.zeros(nSrc, dtype=apdbSchema.column_dtype(columnDef.datatype))
+            if columnDef.nullable:
+                try:
+                    dataInit.fill(None)
+                except (TypeError, ValueError):
+                    pass
```
andy-slac, line 69: For efficiency it may be better to skip all-NULL columns completely. For SQL it will save some CPU and bandwidth by not sending default values. For Cassandra it may be even more important, as Cassandra stores NULLs as "tombstones" and too many tombstones can result in problems.

isullivan: That would actually be helpful for a change I want to make later. Would it be better for those columns to be missing when the tables are supplied to `apdb.store()` in `diaPipe`, or should we provide a list of columns to skip?

andy-slac: I think it would be simpler just to omit them from datframes.

ebellm: Particularly if you are omitting missing (nullable) columns, I think you can do this without looping and creating a bunch of new `pd.Series`.   You just need a dict of `coldef_map = {'column_name': dtype}`

and then you can return `sourcetable.as_type(coldef_map)` which I think should be more efficient.

ebellm: A question which @andy-slac might be able to answer is whether we can instead get the dtypes from `felis.datamodel`, which would allow us to work directly from `sdm_schemas` and not (ab)use the APDB class for type mapping.   Presumably it would make it easier for us to handle the DRP transformations as well.

ebellm (changes requested):
A few small comments and some slightly larger questions about type coercion.

python/lsst/ap/association/utils.py
isullivan: I've pushed a change that duplicates a small amount of code from `dax_apdb`, which separates the SDM type checking from the APDB class.

python/lsst/ap/association/transformDiaSourceCatalog.py
```diff
@@ -140,6 +161,7 @@ def __init__(self, initInputs, **kwargs):
         self.funcs = self.getFunctors()
         self.inputSchema = initInputs['diaSourceSchema'].schema
         self._create_bit_pack_mappings()
+        self.apdbSchema = daxApdb.apdbSchema.ApdbSchema(self.config.schemaFile, self.config.schemaName)
```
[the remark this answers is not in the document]
isullivan: I've changed the implementation so that it is not tied to the APDB any more, and should work for DRP schemas. I have not yet tested that, though.

ebellm (approved):

PR #210, opened 2024-04-26
DM-43416: Migrate AP code to external APDB configs

kfindeisen:
This PR modifies `DiaPipelineTask` to use new-style configs instead of a nested `ApdbConfig` field, and updates the unit tests for `TestTotalUnassociatedObjects` to reflect lsst/verify#124. It also does some simplification of the task's unit tests.

### DIFF
--- a/python/lsst/ap/association/diaPipe.py
+++ b/python/lsst/ap/association/diaPipe.py
@@ -214,9 +214,17 @@ class DiaPipelineConfig(pipeBase.PipelineTaskConfig,
         dtype=str,
         default="deep",
     )
-    apdb = daxApdb.ApdbSql.makeField(
+    apdb = daxApdb.ApdbSql.makeField(  # TODO: remove on DM-43419
         doc="Database connection for storing associated DiaSources and "
             "DiaObjects. Must already be initialized.",
+    )
+    apdb_config_url = pexConfig.Field(
+        dtype=str,
+        default=None,
+        optional=False,
+        doc="A config file specifying the APDB and its connection parameters, "
+            "typically written by the apdb-cli command-line utility. "
+            "The database must already be initialized.",
     )
     validBands = pexConfig.ListField(
         dtype=str,
@@ -293,6 +301,14 @@ class DiaPipelineConfig(pipeBase.PipelineTaskConfig,
             "diaObjects for association.",
     )
     idGenerator = DetectorVisitIdGeneratorConfig.make_field()
+    doConfigureApdb = pexConfig.Field(  # TODO: remove on DM-43419
+        dtype=bool,
+        default=True,
+        doc="Use the deprecated ``apdb`` sub-config to set up the APDB, "
+            "instead of the new config (``apdb_config_url``). This field is "
+            "provided for backward-compatibility ONLY and will be removed "
+            "without notice alongside ``apdb``.",
+    )
 
     def setDefaults(self):
         self.apdb.dia_object_index = "baseline"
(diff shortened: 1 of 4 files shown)


python/lsst/ap/association/diaPipe.py
```diff
@@ -580,7 +607,9 @@ def run(self,
                                    doRunForcedMeasurement=self.config.doRunForcedMeasurement,
                                    )
 
-        return pipeBase.Struct(apdbMarker=self.config.apdb.value,
+        # For historical reasons, apdbMarker is a Config even if it's not meant to be read
```
isullivan, line 624: Minor suggestion for clarity:
`# For historical reasons, apdbMarker is a (empty) Config even if it's not meant to be read`

python/lsst/ap/association/diaPipe.py
```diff
@@ -214,9 +217,12 @@ class DiaPipelineConfig(pipeBase.PipelineTaskConfig,
         dtype=str,
         default="deep",
     )
-    apdb = daxApdb.ApdbSql.makeField(  # TODO: remove on DM-43419
+    apdb = pexConfig.ConfigurableField(  # TODO: remove on DM-43419
+        target=daxApdb.ApdbSql,
```
isullivan: I don't think this is necessary. I can retarget `apdb` to use `ApdbCassandra` without this change, and I don't imagine we'll be using the old system for Cassandra after this is merged.

isullivan (approved):
Looks good, a few minor comments.

python/lsst/ap/association/diaPipe.py
kfindeisen: I don't understand this comment. `target` is a mandatory parameter for `ConfigurableField`; you can't leave it out.

python/lsst/ap/association/diaPipe.py
kfindeisen: I don't see anything historical about emptiness. Maybe "apdbMarker is of storage class Config"?

isullivan: OK, my suggestion was inaccurate. I think it looks surprising to return an empty config, so my request is to add context to this comment explaining why an empty `pexConfig.Config()` is returned.

kfindeisen: Then I'm not sure I understand the question. If `apdbMarker` is a Config, then you must provide _some_ config... or are you just looking for something obvious like "`Config()` is the cheapest valid dataset"?

isullivan: Yes, that's all. Just some clarifying comment about why we are returning an empty config even though everything ran successfully.

python/lsst/ap/association/diaPipe.py
isullivan: To clarify my comment: is this change to support the new config system, or the old one? If it's for the new system then I retract my statement, but if it's to support the old system then I would argue that it is not necessary.

kfindeisen: Can you maybe specify exactly what you mean by "this change"? Surely you're not objecting to the deprecation warning?

isullivan: `apdb = daxApdb.ApdbSql.makeField` -> `apdb = pexConfig.ConfigurableField`

kfindeisen: `makeField` does not support `deprecated`.

isullivan: OK, I didn't know that. Sorry for the noise!

PR #211, opened 2024-05-02
DM-41530: Add individual flag fields to APDB schema

ebellm:


python/lsst/ap/association/trailedSourceFilter.py
```diff
@@ -85,17 +85,21 @@ def run(self, dia_sources, exposure_time):
...
-        flag_map = os.path.join(utils.getPackageDir("ap_association"), "data/association-flag-map.yaml")
-        unpacker = UnpackApdbFlags(flag_map, "DiaSource")
-        flags = unpacker.unpack(dia_sources["flags"], "flags")
+        if "flags" in dia_sources.columns:
+            flag_map = os.path.join(utils.getPackageDir("ap_association"), "data/association-flag-map.yaml")
+            unpacker = UnpackApdbFlags(flag_map, "DiaSource")
+            flags = unpacker.unpack(dia_sources["flags"], "flags")
+            trail_edge_flags = flags["ext_trailedSources_Naive_flag_edge"]
+        else:
+            trail_edge_flags = dia_sources["trail_flag_edge"]
```
parejkoj, line 94: I'm curious why you need the `if` here? We're removing `flags` everywhere, so where could it be coming from? The commit message doesn't say either.

python/lsst/ap/association/transformDiaSourceCatalog.py
```diff
@@ -102,9 +102,10 @@ class TransformDiaSourceCatalogConfig(TransformCatalogBaseConfig,
     )
     doPackFlags = pexConfig.Field(
         dtype=bool,
-        default=True,
+        default=False,
         doc="Do pack the flags into one integer column named 'flags'."
-            "If False, instead produce one boolean column per flag."
+            "If False, instead produce one boolean column per flag.",
+        deprecated="This field is no longer used. Will be removed after v27."
```
parejkoj, line 108: Has to be after v28, sadly, since v27 was tagged a few weeks back.

tests/test_loadDiaCatalogs.py
```diff
@@ -84,7 +84,7 @@ def setUp(self):
 
         # These columns are not in the DPDD, yet do appear in DiaSource.yaml.
         # We don't need to check them against the default APDB schema.
-        self.ignoreColumns = ["band", "bboxSize", "isDipole"]
+        self.ignoreColumns = ["band", "bboxSize", "isDipole", "flags"]
```
parejkoj, line 87: Why not just remove it from DiaSource.yaml?

tests/test_transformDiaSourceCatalog.py
```diff
@@ -175,6 +175,7 @@ def test_computeBBoxSize(self):
     def test_bit_unpacker(self):
         """Test that the integer bit packer is functioning correctly.
         """
+        self.config.doPackFlags = True
```
parejkoj, line 178: Didn't  you deprecate this above?

tests/test_transformDiaSourceCatalog.py
```diff
@@ -175,6 +175,7 @@ def test_computeBBoxSize(self):
     def test_bit_unpacker(self):
```
parejkoj, line 175: Put a TODO with a link to the ticket for the removal of the config, or just remove the test entirely if you've totally removed the functionality (which I think we probably should do on this ticket anyway).

parejkoj (changes requested):
One broad comment (which you may plan to do before taking it out of draft anyway): please flatten most of those commits, since they're doing similar changes to many files.

python/lsst/ap/association/trailedSourceFilter.py
ebellm: My perspective is that we are moving to unpacking the bitpacked flags by default, not removing the capability to do so entirely (until after the deprecation period).  That requires a bit of extra code but isn't a substantial increase in complexity.

tests/test_loadDiaCatalogs.py
ebellm: As above, because it would break the functionality of the bitpacking.

tests/test_transformDiaSourceCatalog.py
ebellm: Yes, but I'm keeping the tests for the now non-default functionality.

parejkoj (approved):

PR #215, opened 2024-05-22
DM-41873: Add max cutout size to packageAlerts

bsmartradio:
For some alerts, the footprint size allows the creation of cutouts which are far too large. This ticket adds a maxCutoutSize config to limit the maximum size. The maximum size will be the maximum trail length allowed for a 30 second exposure in pixels (62.5) plus the expected psf size (40 pixels). This comes to 102 pixels.

### DIFF
--- a/python/lsst/ap/association/packageAlerts.py
+++ b/python/lsst/ap/association/packageAlerts.py
@@ -64,6 +64,17 @@ class PackageAlertsConfig(pexConfig.Config):
         default=30,
         doc="Dimension of the square image cutouts to package in the alert."
     )
+    maxCutoutSize = pexConfig.RangeField(
+        dtype=int,
+        min=0,
+        max=1000,
+        default=102,
+        doc="Dimension of the square image cutouts to package in the alert. The"
+            "default size comes from the max trail length in arcseconds "
+            "(10 deg/day) for a 30 second observation divided by the"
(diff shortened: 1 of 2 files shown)


python/lsst/ap/association/packageAlerts.py
```diff
@@ -64,6 +64,17 @@ class PackageAlertsConfig(pexConfig.Config):
         default=30,
         doc="Dimension of the square image cutouts to package in the alert."
     )
+    maxCutoutSize = pexConfig.RangeField(
+        dtype=int,
+        min=0,
+        max=1000,
+        default=102,
+        doc="Dimension of the square image cutouts to package in the alert. The"
+            "default size comes from the max trail length in arcseconds"
```
ebellm, line 73:
```suggestion
            "default size comes from the max trail length in arcseconds (10 deg/day)"
```

python/lsst/ap/association/packageAlerts.py
```diff
@@ -64,6 +64,17 @@ class PackageAlertsConfig(pexConfig.Config):
...
+            "for a 30 second observation (36000/3600.0/24.0*30) divided by the"
```
ebellm, line 74:
```suggestion
            "for a 30 second observation divided by the"
```

ebellm (approved):
A suggestion to make the docstrings a little clearer but otherwise looks good.

PR #217, opened 2024-05-30
DM-44470 add ra and dec to diaForcedSource table

isullivan:


python/lsst/ap/association/diaForcedSource.py
```diff
@@ -237,9 +239,6 @@ def _calibrate_and_merge(self,
         output_catalog["time_processed"] = DateTime.now().toPython()
         # TODO: propagate actual flags (DM-42355)
 
-        # Drop superfluous columns from output DataFrame.
-        output_catalog.drop(columns=self.config.dropColumns, inplace=True)
```
mrawls, line 241: If you don't drop the columns here, it look like you may wind up with `x` and `y` after all, since they are renamed from `slot_Centroid_x` and `slot_Centroid_y` in this method.

mrawls (commented):
Ideally I would like a unit test that checks the final diaForcedSource column list is exactly what we expect (no missing columns and no extraneous columns). If that's not feasible with a short turnaround, please at minimum run the various ci_packages in Jenkins to ensure sdm_schemas validation passes, which should have the same effect.

python/lsst/ap/association/diaForcedSource.py
isullivan: Note that I still drop the columns, I just moved it out of this method so that `x` and `y` are still available for the check in `_trim_to_exposure`

mrawls: Ah OK - so you drop them later in the process, not earlier - sounds good

mrawls (approved):

isullivan:
Your desired checks on the format for the table are all coming in DM-43856, which will ensure that every table exactly matches the `sdm_schemas` definition.

PR #220, opened 2024-05-31
DM-44535: Express dipoleAngle, trailAngle in position angle on sky instead of detector angle

wmwv:
Use new `ConvertDetectorAngleToPositionAngle` from `pipe_tasks` (in branch of same name) to write out dipoleAngle and trailAngle in position angle on sky (E of N in sky coordinates), instead of orientation (+y from +x in detector coordinates).


isullivan (approved):
Looks good, thank you for fixing up trailAngle as well, that will be needed for solar system processing.

PR #221, opened 2024-06-13
DM-36768: Add psf image to alert cutouts

ebellm:


python/lsst/ap/association/packageAlerts.py
```diff
@@ -389,11 +389,12 @@ def createCcdDataCutout(self, image, skyCenter, extent, photoCalib, srcId):
             CCDData object storing the calibrate information from the input
             difference or template image.
         """
+        point = image.getWcs().skyToPixel(skyCenter)
```
parejkoj, line 392: Wouldn't it be better to pass in the pixel centroid to this method from the diaSource record? No need to do WCS math at all, since we know where the source centroid is.

tests/test_packageAlerts.py
```diff
@@ -265,6 +265,9 @@ def testCreateCcdDataCutout(self):
                                      self.cutoutWcs.wcs.cd)
         self.assertFloatsAlmostEqual(ccdData.data,
                                      calibExposure.getImage().array)
+        self.assertFloatsAlmostEqual(ccdData.psf,
+                                     self.exposure.psf.computeKernelImage(self.center).array)
+        self.assertFloatsAlmostEqual(np.sum(ccdData.psf), 1.0)
```
parejkoj, line 270: Not sure this last one is necessary: you don't need to check that `computeKernelImage` is normalized here, since it's already well tested within afw.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -410,6 +411,8 @@ def createCcdDataCutout(self, image, skyCenter, extent, photoCalib, srcId):
                     % srcId)
             return None
 
+        cutoutPsf = image.psf.computeKernelImage(point).array
```
parejkoj, line 414: Maybe a comment here as to why this is `computeKernelImage` instead of `computeImage`? I'm still not convinced as to which is correct for this use case.

python/lsst/ap/association/packageAlerts.py
ebellm: Can we trust that the (x,y) of the DIASource are the same on the calexp and matched template?

python/lsst/ap/association/packageAlerts.py
ebellm: I'm convinced by [previous discussion](https://lsstc.slack.com/archives/C7ZGQAF25/p1698172795741629) that we want the PSF centered in the array.

python/lsst/ap/association/packageAlerts.py
parejkoj: Yes, those are made to be 1-1 pixel aligned (with the template having a buffer, but the xy0 set such that they are aligned).

parejkoj (commented):
As to the question about the wcs calls further down in `createCcdDataCutout`, please file a ticket and we can look at it: it may or may not be necessary there.

parejkoj (approved):

PR #223, opened 2024-06-14
DM-44850: Add utility to create empty tables

isullivan:


tests/test_diaPipe.py
```diff
@@ -35,6 +35,7 @@
 from lsst.pipe.base.testUtils import assertValidOutput
 
 from lsst.ap.association import DiaPipelineTask
+from lsst.ap.association.utils import make_empty_catalog, convertTableToSdmSchema
```
kfindeisen, line 38: I'd suggest moving this to a `test_utils.py`.

tests/test_diaPipe.py
```diff
@@ -292,6 +293,18 @@ def check_diaObjects(bbox, wcs, diaObjects):
         # Verify that no objects inside the bounding box were removed
         self.assertEqual(nIn1, nIn0)
 
+    def test_make_empty_catalog(self):
+        """Check that an empty catalog has the correct format.
+        """
+        tableName = "DiaObject"
```
kfindeisen, line 299: Does it also work with other tables, e.g. `DiaForcedSource`?

tests/test_diaPipe.py
```diff
@@ -292,6 +293,18 @@ def check_diaObjects(bbox, wcs, diaObjects):
...
+        config = self._makeDefaultConfig(config_file=self.config_file.name)
+        diaPipeTask = DiaPipelineTask(config=config)
+        emptyDiaObjects = make_empty_catalog(diaPipeTask.schema, tableName=tableName)
```
kfindeisen, line 302: If you provide a test schema directly, then there's no need to mix `DiaPipelineTask` into the test...

tests/test_diaPipe.py
```diff
@@ -292,6 +293,18 @@ def check_diaObjects(bbox, wcs, diaObjects):
...
+        emptyDf = pd.DataFrame(columns=["diaObjectId",])
+        emptyDf.set_index("diaObjectId")
+        convertedEmptyDiaObjects = convertTableToSdmSchema(diaPipeTask.schema, emptyDf, tableName=tableName)
+        self.assertSetEqual(set(convertedEmptyDiaObjects.columns), set(emptyDiaObjects.columns))
```
kfindeisen, line 306: If you're comparing two `set`, then `assertSetEqual` is redundant.
```suggestion
        self.assertEqual(set(convertedEmptyDiaObjects.columns), set(emptyDiaObjects.columns))
```

kfindeisen (commented):
I'm worried that `ApdbSchema` is, like `_make_empty_catalog`, an implementation detail that may break at any time. I'd like feedback from @andy-slac before approving.

tests/test_diaPipe.py
isullivan: Yes, it should work with any of the tables. I'll add a `for` loop over a few tables.

tests/test_diaPipe.py
isullivan: That is true, though here I wanted to also test that `diaPipeTask.schema` is getting set correctly. That's why I added this test to `test_diaPipe` instead of creating a `test_utils`, but I am happy to move it out and just make the schema directly.

kfindeisen: I'm not sure that's actually being tested here -- the only thing you compare the schema to is itself.

isullivan: OK, I will move it to `test_utils`, and also add a check of a few of the values.

python/lsst/ap/association/utils.py
```diff
@@ -141,7 +142,7 @@ def convertTableToSdmSchema(apdbSchema, sourceTable, tableName):
 
     Parameters
     ----------
-    apdbSchema : `lsst.dax.apdb.apdbSchema.ApdbSchema`
+    apdbSchema : 'dict' of `lsst.dax.apdb.apdbSchema.ApdbSchema`
```
andy-slac, line 145: This should be `lsst.dax.apdb.schema_model.Table` instead of `lsst.dax.apdb.apdbSchema.ApdbSchema`. Same on the line 128 above.

python/lsst/ap/association/utils.py
```diff
@@ -182,7 +183,7 @@ def dropEmptyColumns(apdbSchema, sourceTable, tableName):
...
```
andy-slac, line 186: `lsst.dax.apdb.schema_model.Table` here too.

python/lsst/ap/association/utils.py
```diff
@@ -196,3 +197,27 @@ def dropEmptyColumns(apdbSchema, sourceTable, tableName):
...
     dropColumns = list(set(nullColNames) & set(nullableList))
     return sourceTable.drop(columns=dropColumns)
+
+
+def make_empty_catalog(apdbSchema, tableName):
+    """Make an empty catalog for a table with a given name.
+
+    Parameters
+    ----------
+    apdbSchema : 'dict' of `lsst.dax.apdb.apdbSchema.ApdbSchema`
```
andy-slac, line 207: And here.

andy-slac (commented):
I agree with Krzysztof's sentiment - `ApdbSchema` presently is not in public API. And I think that you do not actually have to use it, as `Apdb` has a method to return table schema which wraps access to `ApdbSchema`. I'd probably re-implement `readSchemaFromApdb` like this:
```
from lsst.dax.apdb import ApdbTables, schema_model

def readSchemaFromApdb(apdb: Apdb) -> dict[str, schema_model.Table]:
    """...."""
    return {table.name: apdb.tableDef(table) for table in ApdbTables}
```
OTOH, if there is something else in `ApdbSchema` that you want to reuse, we can definitely to extend Apdb's public interface.

isullivan:
Great! Your code snippet for `readSchemaFromApdb` works for my tests, so I have switched over to using it.

python/lsst/ap/association/utils.py
```diff
@@ -124,13 +126,10 @@ def readSchemaFromApdb(apdb):
 
     Returns
     -------
-    schemaTable : 'dict' of `lsst.dax.apdb.apdbSchema.ApdbSchema`
+    schemaTable : dict[str, schema_model.Table | None]
```
kfindeisen, line 129: Would it make more sense to not make an entry instead of returning `None`? I assume that's for tables that aren't in a particular DB instance, but it would be good to document that explicitly.

python/lsst/ap/association/utils.py
```diff
@@ -89,7 +91,7 @@ def readSdmSchemaFile(schemaFile: str,
 
     Returns
     -------
-    schemaTable : 'dict' of `lsst.dax.apdb.apdbSchema.ApdbSchema`
+    schemaTable : 'dict' of `lsst.dax.apdb.schema_model.Table`
```
kfindeisen, line 94:
```suggestion
    schemaTable : `dict` [`str`, `lsst.dax.apdb.schema_model.Table`]
```
Same for all the other places that say "'dict' of" instead of "`dict` of"

tests/test_utils.py
```diff
@@ -0,0 +1,54 @@
...
+from lsst.ap.association.utils import readSchemaFromApdb, make_empty_catalog, convertTableToSdmSchema
+import lsst.dax.apdb as daxApdb
+
+
+class TestUtils(unittest.TestCase):
+
+    def test_make_empty_catalog(self):
+        """Check that an empty catalog has the correct format.
+        """
+        apdb_config = daxApdb.ApdbSql.init_database(db_url="sqlite://")
```
kfindeisen, line 36: IIRC, in-memory APDBs will be removed in a future version. But I think @andy-slac would need to deprecate them first?

tests/test_utils.py
```diff
@@ -0,0 +1,54 @@
...
+        config_file = tempfile.NamedTemporaryFile()
+        self.addCleanup(config_file.close)
+        apdb_config.save(config_file.name)
+        apdb = daxApdb.Apdb.from_uri(config_file.name)
```
kfindeisen, line 40: When writing this kind of code, I recommend reading [the API docs](https://pipelines.lsst.io/v/daily/py-api/lsst.dax.apdb.Apdb.html) instead of cargo-culting:
```suggestion
        apdb = daxApdb.Apdb.from_config(apdb_config)
```

kfindeisen (approved):
Looks good, some documentation and usage comments.

tests/test_utils.py
andy-slac: In-memory SQLite support is already removed since DM-43414. It might work for this particular unit test, but you still should not use it.

kfindeisen: In that case, I recommend using `NamedTemporaryFile` [as a context manager](https://docs.python.org/3.11/library/tempfile.html#tempfile.NamedTemporaryFile) instead of using `addCleanup` (which is more appropriate for stuff created in `setUp`):
```suggestion
        with tempfile.NamedTemporaryFile as apdb_file:
            apdb_config = daxApdb.ApdbSql.init_database(db_url="sqlite://" + apdb_file)
            apdb = daxApdb.Apdb.from_config(apdb_config)
```

isullivan: I was struggling to get your suggestion to work, and realized I only need the schema here, not the full APDB. I will use `readSdmSchemaFile` instead of `readSchemaFromApdb` to get the schema.

PR #224, opened 2024-06-20
DM-44937: DiaForcedSource tables indexed by class, not integer

kfindeisen:
This PR fixes a bug in which `make_empty_catalog` did not create an empty catalog and did not convert the input schema to Pandas datatypes.


isullivan (approved):
Looks good, and I can confirm that the empty catalog has type e.g. `class 'numpy.float64'` before this change and type `float64` with it.

PR #225, opened 2024-06-26
DM-44996: Add error handling to PSF cutouts

isullivan:


python/lsst/ap/association/packageAlerts.py
```diff
@@ -400,29 +401,37 @@ def createCcdDataCutout(self, image, skyCenter, pixelCenter, extent, photoCalib,
...
+            # use image.psf.computeKernelImage to provide PSF centered in the array
+            cutoutPsf = image.psf.computeKernelImage(pixelCenter).array
+        except InvalidParameterError:
+            self.log.warning("Could not calculate PSF for DiaSource with "
+                             "id=%i. InvalidParameterError encountered. Exiting."
+                             % srcId)
+            cutoutPsf = None
+        except InvalidPsfError:
+            self.log.warning("Could not calculate PSF for DiaSource with "
+                             "id=%i. InvalidPsfError encountered. Exiting."
```
mrawls, line 432: Instead of "Exiting," prefer "Returning None for cutoutPsf" (as you noted above for when cutout is set to None)

mrawls (approved):
Approving as a reasonable fix for error handling, but I have reservations about this possibly masking a more fundamental problem with getCutout and pixel vs sky coords. I hope I'm wrong, and either way, the logs this generates about DiaSources with centers in wonky locations will help trace the issue.

PR #226, opened 2024-07-01
DM-45023: Add close to producer

bsmartradio:
Add producer.close() to the alert producer to ensure that the producer closes safely.


erinleighh (approved):
Looks good!

PR #229, opened 2024-07-22
DM-43599: Add progress logs to DiaPipelineTask

kfindeisen:
This PR adds standard status logs to APDB loading, APDB storage, and alert packaging, all steps of `DiaPipelineTask` in which the code may fail to log for a minute or more. The lack of prior logging in PackageAlertsTask means it's hard to tell how much of the silence is from `apdb.store` and how much from packaging.

This PR also adds a loop logger (`utils.logging.PeriodicLogger`) to `PackageAlertsTask`. No such logging is practical for `Apdb` implementations, which treats the DB operations as bulk reads or writes rather than an explicit loop.

### DIFF
--- a/python/lsst/ap/association/packageAlerts.py
+++ b/python/lsst/ap/association/packageAlerts.py
@@ -48,6 +48,7 @@ import lsst.geom as geom
 import lsst.pex.config as pexConfig
 from lsst.pex.exceptions import InvalidParameterError
 import lsst.pipe.base as pipeBase
+import lsst.utils.logging
 from lsst.utils.timer import timeMethod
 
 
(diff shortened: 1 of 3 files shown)


python/lsst/ap/association/packageAlerts.py
```diff
@@ -247,7 +249,14 @@ def run(self,
         diffImPhotoCalib = diffIm.getPhotoCalib()
         calexpPhotoCalib = calexp.getPhotoCalib()
         templatePhotoCalib = template.getPhotoCalib()
+
+        n_sources = len(diaSourceCat)
+        self.log.info("Packaging alerts for %d DiaSources.", n_sources)
+        # Log every 10 seconds as proof of liveness.
+        loop_logger = lsst.utils.logging.PeriodicLogger(self.log, interval=10.0, level=logging.DEBUG)
```
timj, line 256: Using debug here means that the log messages won't turn up for the batch systems to use as proof of liveness. We tend to use VERBOSE log level (which you can get from `self.log.VERBOSE`) for messages that we don't want people to see if they are running locally but we do want batch systems to see (although 10 seconds seems quite fast).

kfindeisen: Indeed, the target client is Prompt Processing (where ideally _any single task_ should run in under 10 seconds, and thus never trigger this logger).

bsmartradio (approved):
Everything here looks good to me.

PR #232, opened 2024-08-20
DM-42606: Add dynamic schema id assignment

bsmartradio:
Schema ids are now assigned based on the schema version

### DIFF
--- a/python/lsst/ap/association/packageAlerts.py
+++ b/python/lsst/ap/association/packageAlerts.py
@@ -368,8 +368,10 @@ class PackageAlertsTask(pipeBase.Task):
             Visit and detector ids of these alerts. Used to write out alerts
             which fail to be sent to the alert stream.
         """
+        schema_id = self.alertSchema.get_schema_id()
+        # Serialize and send alerts with the producer timestamp.
         for alert in alerts:
-            alertBytes = self._serializeAlert(alert, schema=self.alertSchema.definition, schema_id=1)
+            alertBytes = self._serializeAlert(alert, schema=self.alertSchema.definition, schema_id=schema_id)
             try:
                 timestamp = int(time.time() * 1000)  # Current time in milliseconds
                 headers = [("producer_timestamp", str(timestamp).encode('utf-8'))]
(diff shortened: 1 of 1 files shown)


python/lsst/ap/association/packageAlerts.py
```diff
@@ -368,8 +370,13 @@ def produceAlerts(self, alerts, visit, detector):
             Visit and detector ids of these alerts. Used to write out alerts
             which fail to be sent to the alert stream.
         """
+        # Assign a schema id using the schema name.
+        numbers = re.findall(r'\d+', self.alertSchema.definition['name'])
```
ebellm, line 374: As mentioned on the `alert_packet` review, let's use a function defined on `Schema`  here so you can just do `schema_id = self.alertSchema.get_schema_id()` or similar

ebellm (changes requested):

python/lsst/ap/association/packageAlerts.py
```diff
@@ -368,8 +369,11 @@ def produceAlerts(self, alerts, visit, detector):
             Visit and detector ids of these alerts. Used to write out alerts
             which fail to be sent to the alert stream.
         """
+        # Assign a schema id using the schema name.
```
ebellm, line 372:
```suggestion
```

python/lsst/ap/association/packageAlerts.py
```diff
@@ -358,7 +358,8 @@ def createDiaSourceExtent(self, bboxSize):
 
     def produceAlerts(self, alerts, visit, detector):
         """Serialize alerts and send them to the alert stream using
-        confluent_kafka's producer.
+        confluent_kafka's producer. Assign a schema id using the
```
ebellm, line 361:
```suggestion
        confluent_kafka's producer.
```

python/lsst/ap/association/packageAlerts.py
```diff
@@ -358,7 +358,8 @@ def createDiaSourceExtent(self, bboxSize):
...
+        schema name.
```
ebellm, line 362:
```suggestion
```

ebellm (approved):
I don't think we need the comments about assigning the schema id based on name anymore, but otherwise this looks fine.

PR #233, opened 2024-08-31
DM-46050: Add option to use image average PSF for alert cutouts.

isullivan:
Also includes changes from DM-46058, but will be rebased


BrunoSanchez (approved):
Changes look good to me.

PR #236, opened 2024-09-14
DM-46345: Query mpSky with dates in TAI, not UTC

Gerenjie:
Fix a timescale issue

### DIFF
--- a/python/lsst/ap/association/loadDiaCatalogs.py
+++ b/python/lsst/ap/association/loadDiaCatalogs.py
@@ -157,7 +157,7 @@ class LoadDiaCatalogsTask(pipeBase.PipelineTask):
         # Load diaSources and forced sources up to the time of the exposure
         # The timespan may include significant padding, so use the midpoint to
         #  avoid missing valid recent diaSources.
-        visitTime = getMidpointFromTimespan(regionTime.timespan).tai
+        visitTime = getMidpointFromTimespan(regionTime.timespan)
 
         diaSources = self.loadDiaSources(diaObjects, region, visitTime, schema)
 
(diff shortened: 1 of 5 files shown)


isullivan (changes requested):
Please modify `getMidpointFromTimespan` to always return the time in `tai`, and update all of the uses of `getMidpointFromTimespan` in `ap_association`. That function is already documented as returning the midpoint in TAI (which is currently incorrect), and only does not because the use in `mpSkyEphemerisQuery` was going to use it in UTC.

kfindeisen:
That seems a bit odd. The return value is an `astropy.time.Time`, which is not *in* TAI, UTC, or anything else in particular. There's a _default_ scale for if the user doesn't say what representation they want, but a `Time` object itself is simply an abstract point in time. I'd recommend not making any assumptions about what the default scale is, just like @Gerenjie has (not) done.

Gerenjie:
I've modified `getMidpointFromTimespan` so the output is always forced to tai and removed .tai from the ends of `getMidpointFromTimespan` calls everywhere in ap_association. Now just waiting on Jenkins 2485.

isullivan:
Sorry, @Gerenjie I didn't have a chance to respond to Krzysztof's comments yet. My motivation for requesting the change is that the requirements in the [DPDD](https://docushare.lsst.org/docushare/dsweb/Get/LSE-163) state that our time midpoints be in TAI, so I am in favor of making that the default. I appreciate Krzysztof's point about not making assumptions about the time scale, though, and it could be helpful to make it clear that we are choosing TAI in each case by explicitly selecting it. I would like to hear back from @kfindeisen before you merge these changes.

kfindeisen:
Yes, what @isullivan said sounds reasonable to me.

PR #239, opened 2024-09-24
DM-46382: Add number of objects loaded to loadDiaCatalogs logs.

isullivan:


bsmartradio (approved):
Everything looks good!

PR #241, opened 2024-10-04
DM-46614: Add option to filter diaObjects for forced measurement

isullivan:


python/lsst/ap/association/diaForcedSource.py
```diff
@@ -121,7 +128,12 @@ def run(self,
             difference and direct images at DiaObject locations.
         """
 
-        afw_dia_objects = self._convert_from_pandas(dia_objects)
+        nDiaSources = dia_objects.loc[:, "nDiaSources"]
+        objectTable = dia_objects[nDiaSources > self.config.historyThreshold]
```
ebellm, line 132: You could do these two lines more compactly as 

`objectTable = dia_objects.query(f'nDiaSources > {self.config.historyThreshold}')`

python/lsst/ap/association/diaForcedSource.py
```diff
@@ -121,7 +128,12 @@ def run(self,
...
+        if objectTable.empty:
+            return pd.DataFrame()
```
ebellm, line 134: does this need to be wrapped in `self._convert_from_pandas`?

python/lsst/ap/association/diaPipe.py
```diff
@@ -933,3 +928,29 @@ def mergeCatalogs(self, originalCatalog, newCatalog, catalogName):
...
+        """Update the diaObject table with the new diaSource records.
+
+        Parameters
+        ----------
+        diaObjects : `pandas.DataFrame`
+            Table of new DiaObjects merged with their history.
+            This table will be updated in place.
+        diaSources : `pandas.DataFrame`
+            The combined preloaded and associated diaSource catalog.
+        """
```
ebellm, line 942: I think you can do all of this more efficiently with something like

```
nDiaSources = diaSources[['diaObjectId','diaSourceId']].groupby('diaObjectId').agg(len)
nDiaSources.name = "nDiaSources"
# I don't remember if diaObjects is indexed or not.  If it is, we could do this:
del diaObjects["nDiaSources"]
diaObjects = diaObjects.join(nDiaSources, how="left")
```

ebellm (changes requested):
A few suggestions.  Happy to rubber-duck the pandas if needed.

python/lsst/ap/association/diaForcedSource.py
isullivan: No, the dataframe is coerced to the sdm_schemas format in diaPipe. I'll add a comment.

ebellm (approved):

PR #244, opened 2024-10-24
DM-47081: Skip packaging missing forced sources

isullivan:
DiaObjects with too few detections are now skipped for running forced photometry. When packaging alerts, we must also skip packaging those forced sources.

### DIFF
--- a/python/lsst/ap/association/packageAlerts.py
+++ b/python/lsst/ap/association/packageAlerts.py
@@ -210,6 +210,7 @@ class PackageAlertsTask(pipeBase.Task):
             calexp,
             template,
             doRunForcedMeasurement=True,
+            forcedSourceHistoryThreshold=0,
             ):
         """Package DiaSources/Object and exposure data into Avro alerts.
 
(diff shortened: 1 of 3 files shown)


python/lsst/ap/association/packageAlerts.py
```diff
@@ -276,7 +277,7 @@ def run(self,
                 objSourceHistory = diaSrcHistory.loc[srcIndex[0]]
             else:
                 objSourceHistory = None
-            if doRunForcedMeasurement and not diaForcedSources.empty:
+            if doRunForcedMeasurement and diaObject["nDiaSources"] > forcedSourceHistoryThreshold:
```
Gerenjie, line 280: The doc for historyThreshold is `Minimum number of detections of a diaObject required to run forced photometry. Set to 0 to include all diaObjects.` This requires  `diaObject["nDiaSources"] > forcedSourceHistoryThreshold`, which intuitively to me should be >=.

Gerenjie (approved):
Looks good to me pending the discussed change!

PR #245, opened 2024-10-29
DM-47274: Ensure that the DiaObject table has correct format after update

isullivan:
The DiaObject table is relatively small, so an additional check on the format is not too expensive.


BrunoSanchez (approved):
Looks good to me

PR #253, opened 2024-11-21
DM-47736 : Make preloaded_SsObjects optional for ssSingleFrameAssociation

Gerenjie:


python/lsst/ap/association/ssSingleFrameAssociation.py
```diff
@@ -109,7 +110,7 @@ def run(self,
             exposure,
             sourceTable,
             band,
-            solarSystemObjectTable):
+            solarSystemObjectTable=None):
```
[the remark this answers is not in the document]
kfindeisen: Please add entries for `exposure` and `sourceTable` as well, on a separate commit.

python/lsst/ap/association/ssSingleFrameAssociation.py
```diff
@@ -137,8 +138,11 @@ def run(self,
         RuntimeError
             Raised if duplicate DiaObjects or duplicate DiaSources are found.
         """
-        # Associate DiaSources with DiaObjects
-        associatedSsSources = self.associateSources(sourceTable, solarSystemObjectTable, exposure)
+        if solarSystemObjectTable is None:
+            associatedSsSources = None
```
kfindeisen, line 142: It sounds like this still counts as a success for the purposes of pipeline tracking, so I'd rather raise `NoWorkFound` just to be explicit that the task did nothing.

Either way, we'll have to be careful that downstream tasks take `ssSingleFrameAssociatedSources` as an optional input. 😬

kfindeisen (approved):
Looks good, but please make sure the docs are up to date.

python/lsst/ap/association/ssSingleFrameAssociation.py
```diff
@@ -119,10 +120,14 @@ def run(self,
...
         ----------
+        exposure : `lsst.afw.image.ExposureF`
+            Calibrated exposure
+        diaSourceTable : `pandas.DataFrame`
+            Newly detected sources.
         band : `str`
             The band in which the new DiaSources were detected.
         solarSystemObjectTable : `pandas.DataFrame`
-            Preloaded Solar System objects expected to be visible in the image.
+            Preloaded Solar System objects expected to be visible in the image, or None.
```
kfindeisen, line 130:
```suggestion
        solarSystemObjectTable : `pandas.DataFrame` or `None`
            Preloaded Solar System objects expected to be visible in the image.
```

PR #256, opened 2024-12-03
DM-47816: Add metrics to packageAlerts

bsmartradio:


python/lsst/ap/association/packageAlerts.py
```diff
@@ -387,14 +388,17 @@ def produceAlerts(self, alerts, visit, detector):
         """
         schema_id = self.alertSchema.get_schema_id()
         # Serialize and send alerts with the producer timestamp.
+        total_alerts = 0
+        produce_start_timestamp = int(time.time())
```
isullivan, line 392: I don't think you need to convert this to an `int`

python/lsst/ap/association/packageAlerts.py
```diff
@@ -405,6 +409,15 @@ def produceAlerts(self, alerts, visit, detector):
                         f.write(alertBytes)
 
         self.producer.flush()
+        visit_midpoint = Time(alert["diaSource"]["midpointMjdTai"], format='mjd', scale='utc')
```
isullivan, line 412: Shouldn't this be `scale='tai'`? That would mean a difference of 37s, if I recall correctly.
Edit: see below, we should get this time directly from the exposure and not from the source catalog (which derives the time from the exposure).

python/lsst/ap/association/packageAlerts.py
```diff
@@ -405,6 +409,15 @@ def produceAlerts(self, alerts, visit, detector):
...
+        produce_end_timestamp = int(time.time())  # Current time in seconds
+        total_time = produce_end_timestamp - visit_midpoint.unix_tai
+        # Assumed 15 seconds from midpoint and 2 second readout time
+        self.metadata['visit_midpoint'] = visit_midpoint.unix_tai + 15 + 2
```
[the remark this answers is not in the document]
isullivan: Looking through the rest of the task, you should have access to the science image. You should get the exposure midpoint directly from the image, and you can also get the exposure time so that you can calculate the exposure end point. That won't include the 2s readout time, but we can account for that elsewhere for now.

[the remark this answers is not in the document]
isullivan: You can get the exposure midpoint in TAI from the science image with:
`mjd_midpoint = exposure.visitInfo.date.toAstropy().tai.mjd`

python/lsst/ap/association/packageAlerts.py
```diff
@@ -405,6 +409,15 @@ def produceAlerts(self, alerts, visit, detector):
                         f.write(alertBytes)
 
         self.producer.flush()
+        visit_midpoint = Time(alert["diaSource"]["midpointMjdTai"], format='mjd', scale='utc')
+        produce_end_timestamp = int(time.time())  # Current time in seconds
+        total_time = produce_end_timestamp - visit_midpoint.unix_tai
```
isullivan, line 414: Since we compute the `total_time` (or other name, below) metric, we should add a `self.log.info` level log message to report it during processing as well.

isullivan (changes requested):
A number of minor changes, but a major one is that we should get the exposure midpoint and exposure time from the science exposure visitInfo to calculate the timing metric.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -387,14 +388,17 @@ def produceAlerts(self, alerts, visit, detector):
         """
         schema_id = self.alertSchema.get_schema_id()
         # Serialize and send alerts with the producer timestamp.
+        total_alerts = 0
+        produce_start_timestamp = int(time.time())
         for alert in alerts:
             alertBytes = self._serializeAlert(alert, schema=self.alertSchema.definition, schema_id=schema_id)
             try:
-                timestamp = int(time.time() * 1000)  # Current time in milliseconds
+                timestamp = int(time.time() * 1000)  # Current time in seconds
```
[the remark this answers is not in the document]
bsmartradio: Ah, this one wasn't supposed to change. Thanks for catching it.

bsmartradio:
I've made the changes as requested and checked that ap_verify is happy with them. I added the log message, let me know if you want any changes to the message. Just running it with Jenkins now.

bsmartradio:
With the new changes I also had to do a minor update to the unit tests.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -405,6 +410,17 @@ def produceAlerts(self, alerts, visit, detector):
                         f.write(alertBytes)
 
         self.producer.flush()
+        produce_end_timestamp = time.time()  # Current time in seconds
+        total_time = produce_end_timestamp - (midpoint_unix + exposure_time/2.0)
+        # Assumed 15 seconds from midpoint and 2 second readout time
```
isullivan, line 415: Drop the out of date code comment

isullivan (approved):
A few minor formatting requests.

PR #260, opened 2024-12-12
DM-48106: Rename dipole classification "flag" and add attempted dipole classification

abudlong:


andy-slac (approved):
Looks good to me, though I'm not an expert on ap_association.

bsmartradio (approved):
Everything looks good here

PR #261, opened 2024-12-12
DM-48095: AP pipeline does not do solar system association

kfindeisen:
This PR turns on solar system processing by default, as this behavior is useful (or, at worst, does no harm) for AP pipelines in any environment.


Gerenjie (approved):
Looks good!

PR #263, opened 2025-01-07
DM-48279: APDB simulation

isullivan:


python/lsst/ap/association/testApdb.py
```diff
@@ -0,0 +1,575 @@
...
+        doc="Field of view of the camera, in square degrees.",
+    )
+    stellar_density = pexConfig.Field(
+        dtype=float,
+        default=1750,
+        doc="Average number of real transient and variable objects"
+            " detected per square degree. For this simulation, these will"
+            " always be detected, and detected in the same location."
+            "The default is chosen such that:"
+            " Density x Rubin Fov (9.6) x # of visits per night (~600) ~ 10M",
```
andy-slac, line 87: I think LDM-141 specifies average of 1000 visits per night?

python/lsst/ap/association/testApdb.py
```diff
@@ -0,0 +1,575 @@
...
+    )
+    false_positive_ratio = pexConfig.Field(
+        dtype=float,
+        default=4,
+        doc="Average ratio of false detections to real sources."
+            "These will be detected in random locations.",
```
andy-slac, line 93: LDM-141 expects false positives to be 50% of real alerts, if the actual number is 400%, do we need to update LDM-141?

python/lsst/ap/association/testApdb.py
```diff
@@ -0,0 +1,575 @@
...
+        default=None,
+        optional=False,
+        doc="A config file specifying the APDB and its connection parameters, "
+            "typically written by the apdb-cli command-line utility. "
+            "The database must already be initialized.",
+    )
+    survey_area = pexConfig.Field(
+        dtype=float,
+        default=10,
+        doc="Area (in degrees) of the simulated survey",
```
andy-slac, line 73: Hmm, with the area so small the distribution across spatial partitions will be very uneven, that will likely cause uneven load across Cassandra cluster.

python/lsst/ap/association/testApdb.py
```diff
@@ -0,0 +1,575 @@
...
+            Object that generates source IDs and random number generator seeds.
+
+        Returns
+        -------
+        diaForcedSources : `pandas.DataFrame`
+            Catalog of calibrated forced photometered fluxes on both the
+            difference and direct images at DiaObject locations.
+        """
+        # Restrict forced source measurement to objects with sufficient history to be reliable.
+        objectTable = diaObjects.query(f'nDiaSources >= {self.config.historyThreshold}')
```
andy-slac, line 445: Hmm, this depends on `nDiaSources` being correct and that depends on having DiaSource history.

andy-slac (commented):
Few general comments, and I did not read the whole thing.

python/lsst/ap/association/testApdb.py
isullivan: Yes, this was just for testing. The actual value will be 18000.

python/lsst/ap/association/testApdb.py
isullivan: Yes, this is the one place I currently use the diaSource history. I will attempt to re-write it without needing that history, so that I can skip reading diaSources.

python/lsst/ap/association/testApdb.py
isullivan: Good question, since we are nowhere close to 2:1 real to bogus sources currently. This ratio is over a year old, though, and has not been re-evaluated from ComCam data.

PR #266, opened 2025-02-03
DM-48611 : Add additional SSSource columns

Gerenjie:


python/lsst/ap/association/ssoAssociation.py
```diff
@@ -103,24 +105,63 @@ def run(self, diaSourceCatalog, solarSystemObjects, exposure):
...
-            np.polynomial.chebyshev.chebval(ref_time, row['obj_y_poly']),
-            np.polynomial.chebyshev.chebval(ref_time, row['obj_z_poly'])
+            chebval(ref_time, row['obj_x_poly']),
+            chebval(ref_time, row['obj_y_poly']),
+            chebval(ref_time, row['obj_z_poly'])
+        ]), axis=1)
+        solarSystemObjects['obj_velocity'] = solarSystemObjects.apply(lambda row: np.array([
+            chebval(ref_time, Chebyshev(row['obj_x_poly']).deriv().coef),
+            chebval(ref_time, Chebyshev(row['obj_y_poly']).deriv().coef),
+            chebval(ref_time, Chebyshev(row['obj_z_poly']).deriv().coef),
```
mjuric, line 125: Check that the units are correct here (I think km/s make most sense; you know you got it right if you're getting between ~10 and ~25 km/s for typical MBAs).

I don't recall if this version of mpsky normalizes the interpolation parameter (is it rescaled to be in [0, 1)?) -- if so, make sure to take that into account as well.

mjuric (changes requested):
See the one comment about velocities; if that looks OK, go ahead and merge!

python/lsst/ap/association/ssoAssociation.py
```diff
@@ -103,24 +105,63 @@ def run(self, diaSourceCatalog, solarSystemObjects, exposure):
...
         ]), axis=1)
-
         vector = np.vstack(solarSystemObjects['obj_position'].values
                            - solarSystemObjects['obs_position'].values)
         solarSystemObjects[['ra', 'dec']] = np.vstack(hp.vec2ang(vector, lonlat=True)).T
         solarSystemObjects['obs_position_x'], solarSystemObjects['obs_position_y'], \
             solarSystemObjects['obs_position_z'] = np.array(list(solarSystemObjects['obs_position'].values)).T
```
mrawls, line 131: Was this here twice by accident and you are simply deleting duplicate code?

python/lsst/ap/association/ssoAssociation.py
```diff
@@ -103,24 +105,63 @@ def run(self, diaSourceCatalog, solarSystemObjects, exposure):
...
-        solarSystemObjects['obj_position_x'], solarSystemObjects['obj_position_y'], \
-            solarSystemObjects['obj_position_z'] = np.array(list(solarSystemObjects['obj_position'].values)).T
+        solarSystemObjects['heliocentricX'], solarSystemObjects['heliocentricY'], \
+            solarSystemObjects['heliocentricZ'] = np.array(list(solarSystemObjects['obj_position'].values)).T
+        solarSystemObjects['obs_velocity_x'], solarSystemObjects['obs_velocity_y'], \
+            solarSystemObjects['obs_velocity_z'] = np.array(list(solarSystemObjects['obs_velocity'].values)).T
+        solarSystemObjects['heliocentricVX'], solarSystemObjects['heliocentricVY'], \
+            solarSystemObjects['heliocentricVZ'] = np.array(list(solarSystemObjects['obj_velocity'].values)).T
```
mrawls, line 137: It appears you are defining duplicate columns for position AND velocity, which I worry could waste space. Why the duplicated information? Could we have some kind of mapping function that handles alternate column names instead of entire duplicated columns?

python/lsst/ap/association/ssoAssociation.py
```diff
@@ -171,18 +216,26 @@ def run(self, diaSourceCatalog, solarSystemObjects, exposure):
...
-        ssSourceData["expected_dec"] = expected_decs
-        unassociatedObjects = maskedObjects[unAssocObjectMask].drop(columns=['obs_x_poly', 'obs_y_poly',
-                                                                             'obs_z_poly', 'obj_x_poly',
-                                                                             'obj_y_poly', 'obj_z_poly',
-                                                                             'obs_position', 'obj_position',
-                                                                             'associated'])
+        ssSourceData["residualRa"] = residual_ras
+        ssSourceData["residualDec"] = residual_decs
+        ssSourceData[source_column] = dia_ids
+        coords = SkyCoord(ra=ssSourceData['ra'].values * u.deg, dec=ssSourceData['dec'].values * u.deg)
```
mrawls, line 228: Are you certain this will always be given in degrees? It would be safer to retrieve the units in a similar way that you retrieve the values, if possible.

python/lsst/ap/association/ssoAssociation.py
```diff
@@ -171,18 +216,26 @@ def run(self, diaSourceCatalog, solarSystemObjects, exposure):
         self.metadata['nExpectedSsObjects'] = nSolarSystemObjects
         assocSourceMask = diaSourceCatalog["ssObjectId"] != 0
         unAssocObjectMask = np.logical_not(maskedObjects['associated'].values)
-        ssSourceData = pd.DataFrame(ssSourceData, columns=["ssObjectId", "obs_position_x", "obs_position_y",
-                                                           "obs_position_z", "obj_position_x",
-                                                           "obj_position_y", "obj_position_z"])
+        ssSourceData = pd.DataFrame(ssSourceData,
+                                    columns=[
+                                        "ssObjectId", "phaseAngle", "heliocentricDist", "topocentricDist"
+                                    ] + stateVectorColumns)
```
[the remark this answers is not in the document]
Gerenjie: ArrowAstropy could work instead, but this complies with RFC-995 by not input/outputting the dataframe, turning it into an astropy table first.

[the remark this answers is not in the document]
mrawls: While true, much of the motivation behind that RFC was to avoid surprise pandas behaviors that introduce bugs, so if it's not too much trouble, I'd still suggest making the switch.

python/lsst/ap/association/ssoAssociation.py
Gerenjie: The schema asks for separate columns for x, y, and z, but I think lines like
`vector = np.vstack(solarSystemObjects['obj_position'].values
                           - solarSystemObjects['obs_position'].values)`
are more elegant than having three separate subtractions. The duplicate columns are dropped before writing this out, and we don't expect to have more than 100 rows, so the memory usage here should stay very small (unless we find orders of magnitude more asteroids than we thought).

python/lsst/ap/association/ssoAssociation.py
Gerenjie: Units are AU/day, and correct.

PR #279, opened 2025-04-01
DM-49825: Remove UTC timezone from timestamp since pandas will not work with it

timj:


isullivan (approved):
Looks good! I appreciate the detailed comment.

PR #281, opened 2025-04-09
DM-46495: add configurable filter for unassociated diaSources

isullivan:


python/lsst/ap/association/utils.py
```diff
@@ -134,6 +134,30 @@ def readSchemaFromApdb(apdb: Apdb) -> dict[str, schema_model.Table | None]:
...
+        Name of the column to check for in the table.
+    tableName : `str`
+        Name of the table in the schema to use.
+
+    Returns
+    -------
+    `bool`
+        Is the specified column in the schema for the table?
+    """
+    table = apdbSchema[tableName]
```
ebellm, line 154: _might_ be faster to do something like

```
names = [columnDef.name for columnDef table.columns]
for col in colNames:
   if col not in names:
      Raise ValueError
```

where colNames is again a list of everything we're checking.

ebellm (approved):
Minor comments, but marking as approved.

PR #286, opened 2025-04-24
DM-50197: Turn on filtering of unassociated diaSources

isullivan:


python/lsst/ap/association/diaPipe.py
```diff
@@ -673,7 +673,7 @@ def createNewDiaObjects(self, unassociatedDiaSources):
         return pipeBase.Struct(diaSources=unassociatedDiaSources,
                                newDiaObjects=newDiaObjects,
                                nNewDiaObjects=len(newDiaObjects),
-                               marginalDiaSources=Table.from_pandas(marginalDiaSources))
+                               marginalDiaSources=marginalDiaSources)
```
mrawls, line 676: `marginalDiaSources` is still a pandas DataFrame, so I think you need to keep the explicit astropy-ification here if you are expecting an astropy Table as the docstring states. I don't understand what your commit message about breaking the Butler formatter is referring to?

isullivan: I should update the docstring instead. With this conversion in place, the butler.put raises an error: `RuntimeError: Failed to serialize dataset marginal_new_dia_source` , but if it is stored internally as a DataFrame and configured to write to ArrowAstropy as the storageClass then the butler handles the conversion without an error.

PR #289, opened 2025-05-02
DM-50570: Fix trailed source filter

isullivan:
The trailed source length is stored in units of pixels, but the threshold for filtering out long trailed sources is in arcseconds. Add a simple calculation to convert the trail length to the correct units before applying the filter.

### DIFF
--- a/python/lsst/ap/association/filterDiaSourceCatalog.py
+++ b/python/lsst/ap/association/filterDiaSourceCatalog.py
@@ -140,6 +140,12 @@ class FilterDiaSourceCatalogConfig(
             "0.416 arcseconds per second.",
         default=36000/3600.0/24.0,
     )
+    estimatedPixelScale = pexConfig.Field(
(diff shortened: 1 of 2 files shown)


python/lsst/ap/association/filterDiaSourceCatalog.py
```diff
@@ -251,10 +257,57 @@ def _check_dia_source_trail(self, dia_sources, exposure_time):
...
+
+        Parameters
+        ----------
+        catalog : `lsst.afw.table.SourceCatalog`
+            Catalog of sources measured on the difference image.
+
+        Returns
+        -------
+        scale : `float`
+            Pixel scale of the catalog
```
bsmartradio, line 282: Add to documentation that scale is arcseconds/pixel for clarification?

bsmartradio (approved):
Two minor comments, mostly for clarification if someone is reading through this.

PR #294, opened 2025-06-24
DM-50561: Add topic check to make sure incorrect topic is not being used

bsmartradio:


python/lsst/ap/association/packageAlerts.py
```diff
@@ -711,6 +718,11 @@ def _server_check(self):
 
         if not topics:
             raise RuntimeError()
+        if self.kafkaTopic not in topics:
+            raise RuntimeError(
+                f"Topic '{self.kafkaTopic}' not found on the Kafka server."
```
erinleighh, line 723:
```suggestion
                f"Topic '{self.kafkaTopic}' not found on the Kafka server. "
```

python/lsst/ap/association/packageAlerts.py
```diff
@@ -711,6 +718,11 @@ def _server_check(self):
...
+                f"Check that the correct topic has been provided and exists"
```
erinleighh, line 724:
```suggestion
                f"Check that the correct topic has been provided and exists "
```

erinleighh (approved):
Looks good! The only thing I noticed was the multi-line error message needing some spaces.

PR #303, opened 2025-08-12
DM-51970: Filter DIASources with pixelFlags_crCenter.

erinleighh:


python/lsst/ap/association/filterDiaSourceCatalog.py
```diff
@@ -188,6 +194,21 @@ def run(self, diaSourceCat, diffImVisitInfo):
             diaSourceCat = diaSourceCat[~sky_source_column].copy(deep=True)
             self.log.info(f"Filtered {num_sky_sources} sky sources.")
 
+        if self.config.doRemoveCrCenter:
+            crCenter_mask = diaSourceCat["pixelFlags_crCenter"]
```
isullivan, line 198: We may end up wanting to filter multiple flags, so please add a configurable `badFlagList` (or a better name!) and iterate through the flags rather than having hard-coded configs for each.

erinleighh: I was actually going to ask about that in my 1 on 1 with Eric tomorrow! I agree with you, I think that's a much more flexible option. :)

tests/test_filterDiaSourceCatalog.py
```diff
@@ -241,13 +303,21 @@ def test_run_with_all_filters(self):
         filterDiaSourceCatalogTask = FilterDiaSourceCatalogTask(config=self.config)
```
isullivan, line 303: Not very important, but note that the docstring for this test should be updated (there are not 6 sources)

tests/test_filterDiaSourceCatalog.py
```diff
@@ -258,7 +328,7 @@ def test_pixelScale_calculation(self):
         scale = filterDiaSourceCatalogTask._estimate_pixel_scale(self.diaSourceCat)
         # Should be almost but not actually equal
         self.assertNotEqual(self.config.estimatedPixelScale, scale)
-        self.assertAlmostEqual(self.config.estimatedPixelScale, scale)
+        self.assertAlmostEqual(self.config.estimatedPixelScale, scale, places=6)
```
isullivan, line 331: Why add this commit? The tests pass without it.

isullivan (approved):
Looks good! A couple fairly minor comments.

tests/test_filterDiaSourceCatalog.py
erinleighh: It wasn't passing at the default threshold after my changes, though I didn't retry after the most recent changes. I'll run another test!

tests/test_filterDiaSourceCatalog.py
erinleighh: Thanks for spotting this! I also noticed I also forgot to change the docstring for one of the tests I added, so I'll fix that as well.

tests/test_filterDiaSourceCatalog.py
erinleighh: Still failing without that change. :(
`FAILED tests/test_filterDiaSourceCatalog.py::TestFilterDiaSourceCatalogTask::test_pixelScale_calculation - AssertionError: 0.2 != np.float64(0.19999993377350445) within 7 places (np.float64(6.622649556531357e-08) difference)`

PR #304, opened 2025-08-12
DM-51885: Filter glint trails not associated with existing DIAObjects

mrawls:
This PR adds `glint_trail` to the list of flags used in filtering out `unAssociatedSources`.


leeskelvin (approved):
LGTM!

PR #305, opened 2025-08-15
Pass MPCORB throughout

Gerenjie:


python/lsst/ap/association/diaPipe.py
```diff
@@ -647,16 +648,21 @@ def run(self,
...
                                    preloadedDiaSources,
                                    diaForcedSourcesFull,
                                    diffIm,
                                    exposure,
                                    template,
+                                   ssSrc=associatedSsSources,
                                    doRunForcedMeasurement=self.config.doRunForcedMeasurement,
                                    forcedSourceHistoryThreshold=forcedSourceHistoryThreshold,
                                    )
+        if associatedSsSources is not None:
```
ebellm, line 663: I assume you're doing this so the MPCORB columns aren't stored in the butler datatype?  Maybe add a comment to that effect.

python/lsst/ap/association/packageAlerts.py
```diff
@@ -244,6 +246,8 @@ def run(self,
             Calexp used to create the ``diffIm``.
         template : `lsst.afw.image.ExposureF` or `None`
             Template image used to create the ``diffIm``.
+        ssSrc : `astropy.table.Table`
```
ebellm, line 249:
```suggestion
        ssSrc : `astropy.table.Table`, optional
```

python/lsst/ap/association/packageAlerts.py
```diff
@@ -267,28 +272,36 @@ def run(self,
...
         loop_logger = lsst.utils.logging.PeriodicLogger(self.log, interval=10.0)
 
         for srcIndex, diaSource in diaSourceCat.iterrows():
             loop_logger.log("%s/%s sources have been packaged.", len(alerts), n_sources)
-            # Get all diaSources for the associated diaObject.
-            # TODO: DM-31992 skip DiaSources associated with Solar System
-            #       Objects for now.
+
+            alertId = diaSource["diaSourceId"]
             if srcIndex[0] == 0:
```
ebellm, line 287: I believe the diaObjectId is 0 because it defaults that way on construction, and we never change it.  To actually set it to NULL I think we'd have to do it in https://github.com/lsst/pipe_tasks/blob/main/python/lsst/pipe/tasks/ssoAssociation.py.

(Probably ssObjectId is also 0 for non-SSO diaSources as well...)

ebellm (changes requested):
Looks good overall--you'll need a rebase against main as DM-50937 has conflicting changes.

I would like you add a test in both test_packageAlerts and test_diaPipe that exercises the SSSources.

python/lsst/ap/association/diaPipe.py
```diff
@@ -906,34 +919,62 @@ def associateDiaSources(self, diaSourceTable, solarSystemObjectTable, diffIm, di
...
 
         if nullColumns is not None:
             for colName in nullColumns:
-                _setNullColumn(standardizedAssociatedDiaSources, colName)
-        return standardizedAssociatedDiaSources
+                _setNullColumn(standardizedDataFrame, colName)
+        return standardizedDataFrame
+
+    def standardizeTable(self, table, tableName, nullColumns=None):
+        """Convert a catalog to SDM schema format with NULL ID values
```
ebellm, line 955:
```suggestion
        """Convert an astropy table to SDM schema format with NULL ID values
```

python/lsst/ap/association/diaPipe.py
```diff
@@ -906,34 +919,62 @@ def associateDiaSources(self, diaSourceTable, solarSystemObjectTable, diffIm, di
...
+
+        Returns
+        -------
+        standardizedAssociatedDiaSources : pandas.DataFrame
+            The standardized DiaSource catalog
+        """
+        print('standardizing')
+        dataFrame = table.to_pandas()
+        standardizedDataFrame = self.standardizeDataFrame(dataFrame, tableName, nullColumns=nullColumns)
+        print('finished standardizing')
```
ebellm, line 976:
```suggestion
```

PR #306, opened 2025-08-19
DM-52215: Change APDB timestamps to MJD TAI

ebellm:


isullivan (approved):
I think a small helper function might simplify the calls to `DateTime.now`. Also, there were a few places in the unit tests where it looked like the `scale` or `system` was not being set.

isullivan (approved):
Re-reviewed with the unit test additions and workaround for pandas conversions. The new changes look good.

PR #309, opened 2025-08-28
tickets/DM-52333 : Test for possibly-None elements before access

Gerenjie:


python/lsst/ap/association/mpSkyEphemerisQuery.py
```diff
@@ -156,8 +156,9 @@ def run(self, predictedRegionTime):
         mpSkySsObjects, elements = self._mpSkyConeSearch(expCenter, expMidPointEPOCH,
                                                          expRadius + self.config.queryBufferRadiusDegrees,
                                                          mpSkyURL)
-        for element in ELEMENTS_LIST:
-            mpSkySsObjects[self.config.mpcorb_prefix + NAMING_MAP[element]] = elements[element]
+        if elements is not None:
+            for element in ELEMENTS_LIST:
+                mpSkySsObjects[self.config.mpcorb_prefix + NAMING_MAP[element]] = elements[element]
```
kfindeisen, line 161: This works, but it would be more robust to guarantee that `_mpSkyConeSearch` always returns a mapping (I assume? It's not documented...). That way, you don't need a separate check for every future use of `elements`.

A simple way to enforce the guarantee is to have `if elements is None: elements = {}` in the request parsing code.

kfindeisen: Out of scope for this work item, but it seems a bit strange that `_mpSkyConeSearch` returns two "pieces" and then `run` combines them. Is there some deeper significance I'm missing?

kfindeisen (approved):
Looks good, have a suggestion for making the fix a bit sturdier. Please document the second return value of `_mpSkyConeSearch`.

python/lsst/ap/association/mpSkyEphemerisQuery.py
```diff
@@ -228,6 +230,9 @@ def _mpSkyConeSearch(self, expCenter, epochMJD, queryRadius, mpSkyURL):
         mpSkySsObjects : `pandas.DataFrame`
             DataFrame with Solar System Object information and RA/DEC position
             within the visit.
+        elements : `pandas.DataFrame` or `NoneType`
+            Orbital elements of ssObjects (to pass to alerts)
+
```
kfindeisen, line 235:
```suggestion
        elements : `pandas.DataFrame` or `None`
            Orbital elements of ssObjects (to pass to alerts)
```

PR #316, opened 2025-09-16
DM-52493 : Route mpcorb columns around ssSource table standardization

Gerenjie:


python/lsst/ap/association/diaPipe.py
```diff
@@ -624,7 +624,11 @@ def run(self,
 
         associatedSsSources = assocResults.associatedSsSources
         if associatedSsSources is not None:
+            mpcorbColumns = [col for col in associatedSsSources.columns if col[:7] == 'MPCORB_']
+            associatedSsSourceOrbits = associatedSsSources[mpcorbColumns].copy()
             associatedSsSources = self.standardizeTable(associatedSsSources, "SSSource", nullColumns=[])
+            for mpcorbColumn in mpcorbColumns:
```
isullivan, line 630: Instead of adding and later removing columns, could you make a copy of `associatedSsSources` with an appropriate new name and add the columns to that. That new table can be sent to the alert packager, and the unmodified `associatedSsSources` can be written to disk.

isullivan (approved):
See my in-line comment about making a copy of `associatedSsSources`.

PR #317, opened 2025-09-30
DM-52506: Add ROTPA to alert packet cutout headers

bsmartradio:


python/lsst/ap/association/packageAlerts.py
```diff
@@ -325,6 +325,8 @@ def run(self,
                 diffImPhotoCalib,
                 diaSource["diaSourceId"],
                 averagePsf=diffImPsf)
+            diffImCutout.header['ROTPA'] = (diffIm.visitInfo.boresightRotAngle.asRadians(),
```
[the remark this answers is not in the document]
kherner: I don't think there is, but you might have to break it up into multiple lines in the code.

tests/test_packageAlerts.py
```diff
@@ -569,7 +614,12 @@ def testRun_without_produce(self, mock_server_check):
                 pixelPoint,
                 geom.Extent2I(self.cutoutSize, self.cutoutSize),
                 cutout.getPhotoCalib(),
-                1234)
+                1234,
+                cutoutType='science')
```
kherner, line 618: Why isn't the cutout type "difference" here? You're comparing to the stand-in for the difference image, right?

tests/test_packageAlerts.py
```diff
@@ -626,7 +676,8 @@ def testRun_without_produce_use_averagePsf(self, mock_server_check):
...
```
kherner, line 680: Same comment.

python/lsst/ap/association/packageAlerts.py
kherner, line 1: Why not add `cutoutType='template'` after line 345, to be consistent with the others?

python/lsst/ap/association/packageAlerts.py
```diff
@@ -540,6 +544,13 @@ def createCcdDataCutout(self, image, skyCenter, pixelCenter, extent, photoCalib,
                   "cutMinY": cutOutMinY},
             unit=u.nJy)
 
+        if cutoutType in ('difference', 'science'):
+            ccdData.header['ROTPA'] = (image.visitInfo.boresightRotAngle.asDegrees(),
+                                       'Position angle in degrees of the focal plane +Y '
+                                       'with respect to North')
```
kherner, line 550: See my other comment about the docstring length. Maybe it's a total line length problem (I didn't count). As a quick check, try moving the +Y down to the next line.

kherner (approved):

tests/test_packageAlerts.py
bsmartradio: Will swap to difference. I wasn't thinking about that since the result is the same whether or not it is science or difference, but for consistency I will change it to difference.

python/lsst/ap/association/packageAlerts.py
kherner: Would  "Pos angle in deg of the focal plane +Y wrt North" fit?

kherner: It wouldn't, but "Pos angle in deg of focal plane +Y wrt North" just barely fits.

python/lsst/ap/association/packageAlerts.py
bsmartradio: As per Eric's recommendation, I've changed some things so that they all get the ROTPA, even though the template doesn't have one. It will now inherit the one from the science image.

PR #323, opened 2025-11-11
DM-52481: Add MASKPLANE to alert cutout byte stream

BrunoSanchez:
This ticket implements writing to the cutouts in the alert stream the maskplane being set in the ``.flags`` attribute of the CCDData object.

It should be tested for merging to main _after_ https://github.com/astropy/astropy/pull/18862 gets merged and released. Otherwise we won't have the functionality we need, and the alert pipelines will halt when attempting to write the cutouts.


ebellm (changes requested):
Can you add a small test that the mask is propagated in correctly?

BrunoSanchez:
New test added with the serialization roundtrip. Flags get back as they were packed. Only missing metadata is the name of each mask plane bit, which could be included either in the schema or in the documentation.

PR #353, opened 2026-03-31
DM-54316: Remove overrides for deprecated config doRunForcedMeasurement

isullivan:


bsmartradio (approved):
Everything looks good, but there should probably be a unit test which tests if `doRunForcedMeasurement` being turned on does what is expected. Possibly have it only run when running Jenkins?

PR #357, opened 2026-05-11
DM-54878: efficiency improvements for AP association

isullivan:


python/lsst/ap/association/association.py
```diff
@@ -321,27 +321,26 @@ def match(self, dia_objects, dia_sources, score_struct):
...
             used_dia_source[score_idx] = True
             obj_id = score_struct.obj_ids[score_idx]
             associated_dia_object_ids[score_idx] = obj_id
-            dia_sources.loc[score_idx, "diaObjectId"] = obj_id
             n_updated_dia_objects += 1
 
+        # Assign positionally rather than by label — score_idx values from
+        # argsort are 0..N-1, but dia_sources may have a non-contiguous
+        # label index after upstream NaN filtering.
+        dia_sources = dia_sources.copy()
```
ebellm, line 339: why does this need a full copy?

python/lsst/ap/association/association.py
```diff
@@ -321,27 +321,26 @@ def match(self, dia_objects, dia_sources, score_struct):
...
             if used_dia_object[dia_obj_idx]:
                 continue
             used_dia_object[dia_obj_idx] = True
             used_dia_source[score_idx] = True
             obj_id = score_struct.obj_ids[score_idx]
             associated_dia_object_ids[score_idx] = obj_id
-            dia_sources.loc[score_idx, "diaObjectId"] = obj_id
             n_updated_dia_objects += 1
 
+        # Assign positionally rather than by label — score_idx values from
```
ebellm, line 336: I realize we haven't changed the logic in this PR, but this is a bad code smell to me--for all of panda's shortcomings, its index trivially handles filtered rows.

python/lsst/ap/association/association.py
isullivan: It is not strictly needed now, but I didn't want to modify the input `dia_sources` table in place, since that would lead to the `unAssocSSDiaSources` table in `diaPipe` being modified. That is currently never used after association so it would be harmless, but it seemed like it could cause a bug in the future.

python/lsst/ap/association/association.py
isullivan: Good point. This should be easy to keep entirely within Pandas. I'll push a fix.

erinleighh (approved):
Everything looks good to me!

PR #358, opened 2026-05-12
DM-54923: Allow AP association to fall back to second-best match

isullivan:


python/lsst/ap/association/association.py
```diff
@@ -167,72 +199,201 @@ def associate_sources(self, dia_objects, dia_sources):
...
+            return np.where(np.isfinite(sq) & (arr > 0.0),
+                            sq, fallback_sq_deg)
+
+        src_ra = dia_sources["ra"].to_numpy()[src_idx]
+        src_dec = dia_sources["dec"].to_numpy()[src_idx]
+        obj_ra = dia_objects["ra"].to_numpy()[obj_idx]
+        obj_dec = dia_objects["dec"].to_numpy()[obj_idx]
+
+        # Wrap RA difference to (-180, 180].
+        dra = ((src_ra - obj_ra) + 180.0) % 360.0 - 180.0
```
ebellm, line 386: what's the reason to do this offset?  doesn't it have the same boundary problems, just in a different location?  I think we need dra to be correct even at ra=180

python/lsst/ap/association/association.py
```diff
@@ -167,72 +199,201 @@ def associate_sources(self, dia_objects, dia_sources):
...
+        ddec = src_dec - obj_dec
+
+        var_ra = (err_sq(dia_sources, "raErr", src_idx)
+                  + err_sq(dia_objects, "raErr", obj_idx))
+        var_dec = (err_sq(dia_sources, "decErr", src_idx)
+                   + err_sq(dia_objects, "decErr", obj_idx))
+        var_ra = np.maximum(var_ra, sigma_floor_sq_deg)
+        var_dec = np.maximum(var_dec, sigma_floor_sq_deg)
+
+        chi2 = dra*dra/var_ra + ddec*ddec/var_dec
```
ebellm:
```suggestion
        chi2 = dra**2./var_ra + ddec**2./var_dec
```

python/lsst/ap/association/association.py
```diff
@@ -167,72 +199,201 @@ def associate_sources(self, dia_objects, dia_sources):
...
+
+        src_idx = np.repeat(np.arange(n_src, dtype=np.int64), counts)
+        obj_idx = np.fromiter(
+            itertools.chain.from_iterable(candidate_lists),
+            dtype=np.int64, count=n_pairs)
+
+        if (self._has_position_errors(dia_sources)
+                and self._has_position_errors(dia_objects)):
+            scores = self._position_nll(dia_sources, dia_objects, src_idx, obj_idx)
+            # ``max_dist`` is the sole association gate: every candidate
```
ebellm, line 300: I think ultimately we want to replace this with actual probabilistic association (a la NWAY) rather than using a hard max_dist cut.  But no objections for now.

ebellm (approved):
A few minor comments

python/lsst/ap/association/association.py
isullivan: I'll try to make the comment more clear. This adds 180, does the mod 360, then subtracts the 180 offset back off. This makes sure that the difference is never greater than +/-180.

ebellm: Of course!  Sorry, wasn't thinking carefully enough

PR #359, opened 2026-05-15
DM-54962: DIAObject summary stats are null for bands other than the current one

ebellm:


isullivan (approved):
This change looks correct and straightforward, but it would be great if you could also add a unit test to cover multi-band calculations.

PR #362, opened 2026-05-27
DM-54275: Filter diaSources based on bbox when trailed sources unavailable 

bsmartradio:


python/lsst/ap/association/filterDiaSourceCatalog.py
```diff
@@ -210,14 +210,22 @@ def run(self, diaSourceCat, diffImVisitInfo):
...
+            bbox_mask = self._check_dia_source_trail_bbox(diaSourceCat, exposure_time)
+            num_trail_filtered = np.sum(trail_mask & ~bbox_mask)
+            num_bbox_filtered = np.sum(bbox_mask)
+            trail_mask |= bbox_mask
             longTrailedDiaSources = diaSourceCat[trail_mask].copy(deep=True)
             rejectedSources = diaSourceCat[rejected_mask].copy(deep=True)
             rejected_mask |= trail_mask
             diaSourceCat = diaSourceCat[~rejected_mask].copy(deep=True)
 
             self.log.info("%i DiaSources exceed max_trail_length %f arcseconds per second, "
```
mrawls: Suggest breaking this into 2 log messages that only print when num_X_filtered > 0.
In addition,
  
    
      
        Suggested change
      
    
    
      
          
            
                        self.log.info("%i DiaSources exceed max_trail_length %f arcseconds per second, "
          
          
            
                        self.log.info("%i DiaSources exceed max_trail_length (%f arcseconds per second), "

python/lsst/ap/association/filterDiaSourceCatalog.py
```diff
@@ -210,14 +210,22 @@ def run(self, diaSourceCat, diffImVisitInfo):
...
+                          "dropping from source catalog. %i DiaSources had no"
+                          "trail calculation and their bounding box exceeded"
+                          "%f arcseconds in either direction or across the diagonal,"
```
mrawls:
```suggestion
                          "max_trail_length (%f arcseconds per second) in either direction or across the diagonal,"
```

python/lsst/ap/association/filterDiaSourceCatalog.py
```diff
@@ -271,6 +279,45 @@ def _check_dia_source_trail(self, dia_sources, exposure_time):
 
         return trail_mask
 
+    def _check_dia_source_trail_bbox(self, dia_sources, exposure_time):
+        """Check bounding boxes of DiaSources with failed trail measurements.
+
+        For sources where the trail measurement flag is set (indicating the
+        trail length is unreliable), fall back to checking the footprint
```
mrawls: Unreliable or unable to be measured, or both?

python/lsst/ap/association/filterDiaSourceCatalog.py
```diff
@@ -271,6 +279,45 @@ def _check_dia_source_trail(self, dia_sources, exposure_time):
...
+        ----------
+        dia_sources : `lsst.afw.table.SourceCatalog`
+            Input diaSources to check.
+        exposure_time : `float`
+            Exposure time from difference image.
+
+        Returns
+        -------
+        bbox_mask : `numpy.ndarray`
+            Boolean mask for diaSources whose bounding box width, height,
```
mrawls:
```suggestion
            Boolean mask for diaSources whose footprint bounding box width, height,
```

python/lsst/ap/association/filterDiaSourceCatalog.py
```diff
@@ -271,6 +279,45 @@ def _check_dia_source_trail(self, dia_sources, exposure_time):
...
+            or diagonal exceeds the max trail length threshold.
+        """
+        pixelScale = self._estimate_pixel_scale(dia_sources)
```
mrawls, line 318: Could you help me understand why we have to estimate the pixel scale on the fly vs getting it from a more canonically accurate source here (and elsewhere in this task)?

tests/test_filterDiaSourceCatalog.py
```diff
@@ -27,6 +27,8 @@
                                                         FilterDiaSourceReliabilityConfig,
                                                         FilterDiaSourceReliabilityTask)
 import lsst.geom as geom
+import lsst.afw.detection as afwDetect
+import lsst.afw.geom as afwGeom
```
mrawls, line 31: We deprecated most of afwGeom a while back, can you not get what you need without invoking an archaic SpanSet?

python/lsst/ap/association/filterDiaSourceCatalog.py
mrawls: Now that this Task is getting a bit complicated, it's definitely time to please add a full class docstring. See https://developer.lsst.io/python/numpydoc.html#documenting-classes

tests/test_filterDiaSourceCatalog.py
```diff
@@ -342,6 +346,136 @@ def test_pixelScale_calculation(self):
...
+        self.config.doTrailedSourceFilter = True
+        task = FilterDiaSourceCatalogTask(config=self.config)
+        exposure_time = self.visitInfo.getExposureTime()
+        # None of the sources have the trail flag set by default
+        bbox_mask = task._check_dia_source_trail_bbox(self.diaSourceCat, exposure_time)
+        self.assertEqual(np.sum(bbox_mask), 0)
+
+    def test_check_dia_source_trail_bbox_flag_small_bbox(self):
+        """Test sources with ext_trailedSources_Naive_flag set but a small bounding
+        box should are not flagged.
```
mrawls:
```suggestion
        """Test that sources with ext_trailedSources_Naive_flag set but with a small bounding
        box are not flagged.
```

tests/test_filterDiaSourceCatalog.py
```diff
@@ -342,6 +346,136 @@ def test_pixelScale_calculation(self):
...
+        task = FilterDiaSourceCatalogTask(config=self.config)
+        exposure_time = self.visitInfo.getExposureTime()
+        # Set the trail flag on a source with a small footprint (PSF-sized)
+        self.diaSourceCat[0]["ext_trailedSources_Naive_flag"] = True
+        bbox_mask = task._check_dia_source_trail_bbox(self.diaSourceCat, exposure_time)
+        # The default PSF footprint is small (~7 pixels), well below the
+        # threshold for max_trail_length * exposure_time / pixelScale
+        self.assertFalse(bbox_mask[0])
+
+    def test_check_dia_source_trail_bbox_flag_large_bbox(self):
```
mrawls, line 427: Is it ever possible to have ext_trailedSources_Naive_flag *not* set and to have a bbox measured? (asking as a possible other unit test to run)

python/lsst/ap/association/filterDiaSourceCatalog.py
```diff
@@ -210,14 +210,22 @@ def run(self, diaSourceCat, diffImVisitInfo):
 
         if self.config.doTrailedSourceFilter:
             trail_mask = self._check_dia_source_trail(diaSourceCat, exposure_time)
+            bbox_mask = self._check_dia_source_trail_bbox(diaSourceCat, exposure_time)
```
mrawls, line 221: Would it be possible to only run this if the initial trail measurement fails? It looks like you calculate it every time, but perhaps there's a reason for that.

mrawls (commented):
I'm convinced this does what you want, but it's necessary to get pretty deep in the weeds to figure out what it is doing. My main suggestion is to please add a detailed class docstring that explains all the various ways diaSources are filtered and what the resulting flag situation looks like in each case.

python/lsst/ap/association/filterDiaSourceCatalog.py
bsmartradio: I meant to type unavailable, but I think swapping it to unable to be measured is the most clear.

python/lsst/ap/association/filterDiaSourceCatalog.py
bsmartradio: We do, but only in the function. Doing that here would maybe be a bit messy because I'm sending the whole source catalog. Instead I've done an initial filter inside the function so that it doesn't bother checking each source individually. I probably should have done it that way to begin with for speed.

tests/test_filterDiaSourceCatalog.py
bsmartradio: I added one more unit check that checks that sources with no flag but large bboxes do not get filtered

python/lsst/ap/association/filterDiaSourceCatalog.py
mrawls: That makes more sense - yes, I like "unable to be measured" best. I don't think you've implemented that change yet.

python/lsst/ap/association/filterDiaSourceCatalog.py
mrawls: Per offline conversation, it does seem this is necessary, since you only have a source catalog here, not an image. All good.

tests/test_filterDiaSourceCatalog.py
mrawls: Per offline conversation, it does seem that this is where SpanSets need to come from, and you're using "regular" geom for everything else. All good.

python/lsst/ap/association/filterDiaSourceCatalog.py
```diff
@@ -169,6 +171,10 @@ class FilterDiaSourceCatalogTask(pipeBase.PipelineTask):
     def run(self, diaSourceCat, diffImVisitInfo):
         """Filter sources from the supplied DiaSource catalog.
 
+        Sources are filtered based on their trail length, sky sources, and bad
+         flags. Aby source which meets the filtering criteria is removed from
```
mrawls: There's a typo (Aby --> Any) and also some odd indentation happening here.

Suggested rephrase: "DiaSources are filtered based on criteria including trail length, whether they are a sky source, and whether they have one or more bad flags. Any source which meets a filtering criteria is removed from the main output catalog and optionally saved in one or more separate output catalogs."

mrawls (approved):

